PR #36, opened 2020-01-23
Mc samples

f-dangel:
Allow choice of MC samples for `KFAC` and `DiagGGNMC`


backpack/core/derivatives/crossentropyloss.py
```diff
@@ -26,8 +26,14 @@ def sqrt_hessian(self, module, g_inp, g_out):
 
         return sqrt_H
 
-    def sqrt_hessian_sampled(self, module, g_inp, g_out):
-        M = self.MC_SAMPLES
+    def make_sqrt_hessian_sampled_fn(self, mc_samples=1):
+        def _sqrt_hessian_sampled_fn(module, g_inp, g_out):
+            return self.sqrt_hessian_sampled(module, g_inp, g_out, mc_samples)
+
+        return _sqrt_hessian_sampled_fn
```
fKunstner: This function seems overly complex. A [`partial evaluation`](https://docs.python.org/3/library/functools.html#functools.partial) for the specialization of the `mc_sample` parameter seems simpler.
```
def make_sqrt_hessian_sampled_fn(self, mc_samples=1):
    return partial(self.sqrt_hessian_sampled, mc_samples=mc_samples)
````

File: backpack/core/derivatives/mseloss.py [hunk 2]
```diff
@@ -21,7 +21,18 @@ def sqrt_hessian(self, module, g_inp, g_out):
...
+        def _sqrt_hessian_sampled_fn(module, g_inp, g_out):
+            return self.sqrt_hessian_sampled(module, g_inp, g_out, mc_samples)
+
+        return _sqrt_hessian_sampled_fn
+
+    def sqrt_hessian_sampled(self, module, g_inp, g_out, mc_samples=None):
+        """
+        Note:
+        -----
+        Parameter `mc_samples` will be ignored.
```
fKunstner: Can you add the reason why `mc_samples` is ignored?

backpack/extensions/secondorder/diag_ggn/losses.py
```diff
@@ -7,17 +7,22 @@
...
+
+    def make_loss_hessian_func(self, ext):
+        """Get function that produces the backpropagated quantity."""
+        loss_hessian_strategy = ext.loss_hessian_strategy
+
+        if loss_hessian_strategy == LossHessianStrategy.EXACT:
+            return self.derivatives.sqrt_hessian
+        elif loss_hessian_strategy == LossHessianStrategy.SAMPLING:
+            mc_samples = ext.get_num_mc_samples()
+            return self.derivatives.make_sqrt_hessian_sampled_fn(mc_samples)
```
fKunstner: You can also make the partial evaluation here directly, rather than use `make_sqrt_hessian_sampled_fn`.
```
return partial(self.sqrt_hessian_sampled, mc_samples=mc_samples)
```

backpack/extensions/secondorder/hbp/losses.py
```diff
@@ -6,26 +6,30 @@
...
 
+    def make_loss_hessian_func(self, ext):
+        """Get function that produces the backpropagated quantity."""
+        hessian_strategy = ext.get_loss_hessian_strategy()
+
+        if hessian_strategy == LossHessianStrategy.EXACT:
+            return self.derivatives.sqrt_hessian
+        elif hessian_strategy == LossHessianStrategy.SAMPLING:
+            mc_samples = ext.get_num_mc_samples()
+            return self.derivatives.make_sqrt_hessian_sampled_fn(mc_samples)
```
fKunstner: Same as above, use `partial`?

backpack/extensions/secondorder/hbp/losses.py
```diff
@@ -6,26 +6,30 @@
...
+        elif hessian_strategy == LossHessianStrategy.AVERAGE:
+            return self.derivatives.sum_hessian
```
fKunstner: Is there a naming inconsistency here? Why does the `AVERAGE` strategy calls the `sum_*`?

f-dangel: It should be `SUM`

PR #37, opened 2020-01-24
Core: Change tensor shape conventions

f-dangel:
This PR changes the internal shape conventions of `backpack` without changing the API.

**Most important changes:**
* Vectorization axis for Jacobian-matrix products moves from `-1` to `0` and feature dimensions are not flattened anymore, saving some `view`s and `contiguous`s. 

Example: Consider The `jac_t_mat_prod` for a `Conv2d` layer with input of shape `[N, C_in, H_in, W_in]` and output of shape `[N, C_out, H_out, W_out]`. 
**old convention:**   `[N, C_out * H_out * W_out, V]` -> `[N, C_in * H_in * W_in]`
**new convention:** `[V, N, C_out, H_out, W_out]` -> `[V, N, C_in, H_in, W_in]`
(`V` denoting the vectorization axis)

Background: Most of the routines looked like that
```python
def jac_mat_prod(..., mat):
    1) structure = get_structure_from_io()
    2) mat = recover_feature_structure(mat, structure)
    3) Jmat = do_the_actual_jac_mat(mat)
    4) Jmat = flatten_feature_structure(mat, structure)
    5) return Jmat   
```
In most cases, steps 1), 2), 4) were eliminated or significantly shortened.

* This implies that the `Flatten` module has to perform operations to do a Jacobian-matrix product. It is now not a no-op anymore.

* Deprecation of `LinearConcat` and `Conv2dConcat`. The concatenated layer cannot be integrated in the new convention, and we did not have a good use case for them.

* Since the vectorization axis moved, the interface to the `curvmatprod` library changed, which is not problematic as it's not released yet. Modified the tests.

**Refactoring:**
* Add shape checking to the Jacobian-matrix routines
* Refactor the `BaseDerivatives` class, pulling out functions for parametrized layers and loss layers into `BaseParameterDerivatives` and `BaseLossDerivatives`
* Improve naming of axes, closer to the `torch` convention
* Replace some relative imports with absolute ones

**Helpers:**
* [`try_view`](https://github.com/f-dangel/backpack/blob/new-shape-convention/backpack/utils/einsum.py#L102-L107): Try to call `view`, fall back to `reshape` if contiguous is required
* [`eingroup`](https://github.com/f-dangel/backpack/blob/new-shape-convention/backpack/utils/einsum.py#L41-L56): convenience function for index (un)grouping (needs more documentation)


backpack/utils/einsum.py
```diff
@@ -34,3 +36,72 @@ def einsum(equation, *operands):
...
+
+        for group, size in zip(in_groups, operand.shape):
+            if len(group) == 1:
+                axis = group[0]
+                if axis in dim.keys():
+                    print("[eingroup] Can infer dim of input group {}".format(group))
+                    assert dim[axis] == size
+                dim[axis] = size
+            else:
+                print("[eingroup] Cannot infer dim of input group {}".format(group))
```
fKunstner: Remove print statements, throw error on failure instead.

PR #42, opened 2020-02-10
Lint

fKunstner:


backpack/extensions/curvmatprod/losses.py
```diff
@@ -1,8 +1,9 @@
-from backpack.core.derivatives.mseloss import MSELossDerivatives
 from backpack.core.derivatives.crossentropyloss import CrossEntropyLossDerivatives
-from .cmpbase import CMPBase
+from backpack.core.derivatives.mseloss import MSELossDerivatives
 from backpack.extensions.curvature import Curvature
 
+from .cmpbase import CMPBase
```
f-dangel: Could you replace this by an absolute import for consistency?

PR #100, opened 2020-08-17
Batch l2 gradients: extensions 

sbharadwajj:
1.  BatchL2Grad extensions for `Conv1d`, `Conv3d`, `ConvTranspose1d`, `ConvTranspose2d`, `ConvTranspose3d`
2. Test suite for the same

### DIFF
--- a/backpack/utils/conv_transpose.py
+++ b/backpack/utils/conv_transpose.py
@@ -9,9 +9,7 @@ def get_convtranspose1d_weight_gradient_factors(input, grad_out, module):
     kernel_size = module.kernel_size
     kernel_size_numel = int(torch.prod(torch.Tensor(kernel_size)))
 
-    X = unfold_by_conv_transpose(input, module).reshape(
-        N, C_in * kernel_size_numel, -1
-    )
+    X = unfold_by_conv_transpose(input, module).reshape(N, C_in * kernel_size_numel, -1)
     return X, grad_out
 
 
@@ -20,9 +18,7 @@ def get_convtranspose2d_weight_gradient_factors(input, grad_out, module):
     kernel_size = module.kernel_size
     kernel_size_numel = int(torch.prod(torch.Tensor(kernel_size)))
 
-    X = unfold_by_conv_transpose(input, module).reshape(
-        N, C_in * kernel_size_numel, -1
-    )
+    X = unfold_by_conv_transpose(input, module).reshape(N, C_in * kernel_size_numel, -1)
     dE_dY = eingroup("n,c,h,w->n,c,hw", grad_out)
     return X, dE_dY
 
@@ -32,9 +28,7 @@ def get_convtranspose3d_weight_gradient_factors(input, grad_out, module):
     kernel_size = module.kernel_size
     kernel_size_numel = int(torch.prod(torch.Tensor(kernel_size)))
 
-    X = unfold_by_conv_transpose(input, module).reshape(
-        N, C_in * kernel_size_numel, -1
-    )
+    X = unfold_by_conv_transpose(input, module).reshape(N, C_in * kernel_size_numel, -1)
     dE_dY = eingroup("n,c,d,h,w->n,c,dhw", grad_out)
     return X, dE_dY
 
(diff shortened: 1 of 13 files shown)


backpack/utils/conv_transpose.py
```diff
@@ -1,6 +1,44 @@
...
 
+from backpack.utils.ein import eingroup
+
+
+def get_convtranspose1d_weight_gradient_factors(input, grad_out, module):
+    N, C_in = input.shape[0], input.shape[1]
+    kernel_size = module.kernel_size
+    kernel_size_numel = int(torch.prod(torch.Tensor(kernel_size)))
+
+    X = (torch.flatten(unfold_by_conv_transpose(input, module))).reshape(
```
f-dangel, line 12: The `torch.flatten(...)` should not be required. Please check and remove it if possible

backpack/utils/conv_transpose.py
```diff
@@ -1,6 +1,44 @@
...
+    dE_dY = eingroup("n,c,h,w->n,c,hw", grad_out)
+    return X, dE_dY
+
+
+def get_convtranspose3d_weight_gradient_factors(input, grad_out, module):
+    N, C_in = input.shape[0], input.shape[1]
+    kernel_size = module.kernel_size
+    kernel_size_numel = int(torch.prod(torch.Tensor(kernel_size)))
+
+    X = (torch.flatten(unfold_by_conv_transpose(input, module))).reshape(
```
f-dangel, line 36: The `torch.flatten(...)` should not be required. Please check and remove it if possible

backpack/utils/conv_transpose.py
```diff
@@ -1,6 +1,44 @@
...
+    dE_dY = eingroup("n,c,l->n,c,l", grad_out)
+    return X, dE_dY
+
+
+def get_convtranspose2d_weight_gradient_factors(input, grad_out, module):
+    N, C_in = input.shape[0], input.shape[1]
+    kernel_size = module.kernel_size
+    kernel_size_numel = int(torch.prod(torch.Tensor(kernel_size)))
+
+    X = (torch.flatten(unfold_by_conv_transpose(input, module))).reshape(
```
f-dangel, line 24: The `torch.flatten(...)` should not be required. Please check and remove it if possible

backpack/utils/conv_transpose.py
```diff
@@ -1,6 +1,44 @@
...
+
+def get_convtranspose1d_weight_gradient_factors(input, grad_out, module):
+    N, C_in = input.shape[0], input.shape[1]
+    kernel_size = module.kernel_size
+    kernel_size_numel = int(torch.prod(torch.Tensor(kernel_size)))
+
+    X = (torch.flatten(unfold_by_conv_transpose(input, module))).reshape(
+        N, C_in * kernel_size_numel, -1
+    )
+    dE_dY = eingroup("n,c,l->n,c,l", grad_out)
```
f-dangel, line 15: That call seems to do nothing. Please check and remove if possible

test/extensions/implementation/autograd.py
```diff
@@ -29,3 +29,10 @@ def batch_grad(self):
                 batch_grads[idx][b, :] = g.detach() * factor
 
         return batch_grads
+
+    def batch_l2_grad(self):
+        batch_grad = self.batch_grad()
+        batch_l2_grads = [
+            (g ** 2).sum(list(range(1, len(g.shape)))) for g in batch_grad
```
f-dangel, line 36: The summation over all axes except the `0`th could also be written as `(g ** 2).flatten(start_dim=1).sum(1) for g in ...`. Do you think that this way the code would be easier to read?

backpack/utils/conv_transpose.py
sbharadwajj: yes. But I suggest we still return `grad_out` as `dE_dY` to maintain uniformity across other functions. What do you suggest?

test/extensions/implementation/autograd.py
sbharadwajj: Yes, this is a better way to write it

PR #106, opened 2020-09-02
Refactor SumGrad 

sbharadwajj:
ADD: refactored `SumGrad` by adding base class

### DIFF
--- a/backpack/extensions/firstorder/sum_grad_squared/conv1d.py
+++ b/backpack/extensions/firstorder/sum_grad_squared/conv1d.py
@@ -1,7 +1,7 @@
 from backpack.core.derivatives.conv1d import Conv1DDerivatives
-from backpack.extensions.firstorder.sum_grad_squared.sum_grad_base import SumGradBase
+from backpack.extensions.firstorder.sum_grad_squared.sgs_base import SGSBase
 
 
-class SGSConv1d(SumGradBase):
+class SGSConv1d(SGSBase):
     def __init__(self):
-        super().__init__(derivatives=Conv1DDerivatives, params=["bias", "weight"])
+        super().__init__(derivatives=Conv1DDerivatives(), params=["bias", "weight"])
--- a/backpack/extensions/firstorder/sum_grad_squared/conv1d.py
+++ b/backpack/extensions/firstorder/sum_grad_squared/conv1d.py
@@ -1,21 +1,7 @@
-from torch import einsum
-
-from backpack.extensions.firstorder.base import FirstOrderModuleExtension
-from backpack.utils import conv as convUtils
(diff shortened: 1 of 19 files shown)


backpack/extensions/firstorder/sum_grad_squared/conv1d.py
```diff
@@ -1,21 +1,7 @@
-from torch import einsum
+from backpack.core.derivatives.conv1d import Conv1DDerivatives
+from backpack.extensions.firstorder.sum_grad_squared.sum_grad_base import SumGradBase
```
f-dangel, line 2: Just a minor thing: Can you rename `SumGradBase` into `SGSBase` and `sum_grad_base.py` into `sgs_base.py`

backpack/extensions/firstorder/sum_grad_squared/sum_grad_base.py
```diff
@@ -0,0 +1,20 @@
+from backpack.extensions.firstorder.base import FirstOrderModuleExtension
+
+
+class SumGradBase(FirstOrderModuleExtension):
+    def __init__(self, derivatives, params=None):
+        self.derivatives = derivatives()
```
f-dangel, line 6: Can you make this more consistent with other first-order conventions, where the base class accepts an instance of a derivative, and not the object itself ([example](https://github.com/f-dangel/backpack/blob/master/backpack/extensions/firstorder/batch_grad/batch_grad_base.py).

It should be `self.derivatives = derivatives`, and the constructor should be called in the `__init__` of a child class ([example](https://github.com/f-dangel/backpack/blob/master/backpack/extensions/firstorder/batch_grad/linear.py)).

backpack/extensions/firstorder/sum_grad_squared/sum_grad_base.py
```diff
@@ -0,0 +1,20 @@
...
+        grad_batch = self.derivatives.bias_jac_t_mat_prod(
+            module, g_inp, g_out, g_out[0], sum_batch=False
+        )
+        return (grad_batch ** 2).sum(self.N_axis)
+
+    def weight(self, ext, module, g_inp, g_out, bpQuantities):
+        grad_batch = self.derivatives.weight_jac_t_mat_prod(
+            module, g_inp, g_out, g_out[0], sum_batch=False
+        )
+        return (grad_batch ** 2).sum(self.N_axis).view_as(module.weight)
```
f-dangel, line 20: Is the `view_as(module.weight)` necessary?

sbharadwajj: surprisingly not. Sorry when I had tried it for the first time it threw a `shape error`.

PR #107, opened 2020-09-05
Change test core settings to new format

sbharadwajj:
1. Initial test pull request to check if old and new tests pass
2. old tests have not been removed yet

### DIFF
--- a/test/core/derivatives/problem.py
+++ b/test/core/derivatives/problem.py
@@ -16,10 +16,7 @@ def make_test_problems(settings):
     problem_dicts = []
 
     for setting in settings:
-        if is_new_format(setting):
-            setting = add_missing_defaults_new_format(setting)
-        else:
-            setting = add_missing_defaults_old_format(setting)
+        setting = add_missing_defaults(setting)
         devices = setting["device"]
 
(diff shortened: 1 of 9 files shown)


sbharadwajj:
3. Removal of the old test format
4. change documentation

Note: Test format for 'utils/test_conv_settings.py` and `utils/test_conv_transpose_settings.py` was also changed to match the new format because we imported `problem.py` there aswell.

test/core/derivatives/problem.py
```diff
@@ -28,22 +28,18 @@ def make_test_problems(settings):
...
+        "target_fn": lambda: None,
         "id_prefix": "",
+        "device": get_available_devices(),
         "seed": 0,
-        "target_fn": lambda: None,
+    }
+    overwrite = {
+        "module_kwargs": {},
+        "input_kwargs": {},
         "target_kwargs": {},
```
f-dangel, line 42: The arguments `module_kwargs`, `input_kwargs`, and `target_kwargs` are legacy from the old format,not required by the new format. They should be removed. You will also have to refactor the implementation of `DerivativesTestProblem` below for that to work.

PR #111, opened 2020-10-22
ADD: refactord batch-l2

sbharadwajj:
refactored `batch-l2-grad` to a single base class

### DIFF
--- a/backpack/extensions/firstorder/batch_l2_grad/conv1d.py
+++ b/backpack/extensions/firstorder/batch_l2_grad/conv1d.py
@@ -1,6 +1,6 @@
-from backpack.extensions.firstorder.batch_l2_grad.conv_base import BatchL2ConvBase
+from backpack.extensions.firstorder.batch_l2_grad.convnd import BatchL2Convnd
 
 
-class BatchL2Conv1d(BatchL2ConvBase):
+class BatchL2Conv1d(BatchL2Convnd):
     def __init__(self):
-        super().__init__(N=1, params=["bias", "weight"], convtranspose=False)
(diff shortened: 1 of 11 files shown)


backpack/extensions/firstorder/batch_l2_grad/conv1d.py
```diff
@@ -1,19 +1,6 @@
-from torch import einsum
+from backpack.extensions.firstorder.batch_l2_grad.convnd import BatchL2Convnd
```
f-dangel, line 1: Please improve consistency with the [`core`](https://github.com/f-dangel/backpack/blob/master/backpack/core/derivatives/convnd.py#L13) by renaming `BatchL2Convnd` into `BatchL2ConvNd`

backpack/utils/conv.py
```diff
@@ -15,24 +15,32 @@ def unfold_func(module):
...
+        dE_dY = eingroup("n,c,h,w->n,c,hw", grad_out)
+    elif N == 3:
+        dE_dY = eingroup("n,c,d,h,w->n,c,dhw", grad_out)
+    else:
+        raise ValueError("{}-dimensional Conv. is not implemented.".format(N))
 
-def get_weight_gradient_factors(input, grad_out, module):
-    # shape [N, C_in * K_x * K_y, H_out * W_out]
-    X = unfold_func(module)(input)
-    dE_dY = eingroup("n,c,h,w->n,c,hw", grad_out)
```
f-dangel, line 27: For `N==2`, `unfold_func` should be used as it uses `torch.nn.Unfold`, which is faster than our `unfold_by_conv` generalization to 1 and 3 dimensions.

PR #115, opened 2020-10-25
Diagh: extended

sbharadwajj:
Note: `diag_ggn/conv2d.py` is changed only so that the tests pass

Must be reviewed only after the other pull requests


backpack/extensions/secondorder/diag_hessian/activations.py
```diff
@@ -2,6 +2,11 @@
 from backpack.core.derivatives.sigmoid import SigmoidDerivatives
 from backpack.core.derivatives.tanh import TanhDerivatives
 from backpack.extensions.secondorder.diag_hessian.diag_h_base import DiagHBaseModule
+from backpack.core.derivatives.elu import ELUDerivatives
+from backpack.core.derivatives.selu import SELUDerivatives
+from backpack.core.derivatives.leakyrelu import LeakyReLUDerivatives
+from backpack.core.derivatives.logsigmoid import LogSigmoidDerivatives
+from backpack.extensions.secondorder.diag_ggn.diag_ggn_base import DiagGGNBaseModule
```
f-dangel, line 9: I think that is not correct, you are using the `DiagGGNBaseModule` but it should be the `DiagHBaseModule`

backpack/extensions/secondorder/diag_hessian/convtransposend.py
```diff
@@ -0,0 +1,32 @@
+import torch
+import torch.nn
+
+from backpack.extensions.secondorder.diag_hessian.diag_h_base import DiagHBaseModule
+from backpack.utils import conv_transpose as convUtils
+
+
+class DiagHConvTransposeND(DiagHBaseModule):
+    def __init__(self, derivatives, N, params=None, convtranspose=False):
```
f-dangel, line 9: The `convtranspose` argument is not used and can be removed

backpack/utils/conv_transpose.py
```diff
@@ -33,6 +34,56 @@ def get_convtranspose3d_weight_gradient_factors(input, grad_out, module):
...
+
+    def get_output_shape(input, module):
+        return module(input).shape
+
+    out_channels = module.weight.shape[0]
+    in_channels = module.weight.shape[1]
+    k = module.weight.shape[2:]
+
+    M = input.shape[0]
+    output_shape = get_output_shape(module.input0, module)
```
f-dangel, line 51: `get_output_shape` is only used once, let's get rid of this function and inline it

backpack/extensions/secondorder/diag_hessian/activations.py
```diff
@@ -17,3 +22,23 @@ def __init__(self):
...
+
+
+class DiagGGNLeakyReLU(DiagHBaseModule):
+    def __init__(self):
+        super().__init__(derivatives=LeakyReLUDerivatives())
+
+
+class DiagGGNLogSigmoid(DiagHBaseModule):
+    def __init__(self):
+        super().__init__(derivatives=LogSigmoidDerivatives())
```
f-dangel, line 44: Replace `DiagGGN` with `DiagH`

sbharadwajj: sorryyyy

test/extensions/secondorder/secondorder_settings.py
```diff
@@ -70,6 +70,63 @@
...
+        "input_fn": lambda: torch.rand(3, 3, 7),
+        "module_fn": lambda: torch.nn.Sequential(
+            torch.nn.Conv1d(3, 2, 2, bias=False),
+            torch.nn.ReLU(),
+            torch.nn.Flatten(),
+            torch.nn.Linear(12, 5),
+        ),
+        "loss_function_fn": lambda: torch.nn.CrossEntropyLoss(reduction="sum"),
+        "target_fn": lambda: classification_targets((3,), 5),
+    },
```
f-dangel, line 94: Meta-comment: One pro of the new test suite is that it is easy to add new tests explicitely. Now we are at a stage where it is hard to me to figure out the test coverage by staring at the big dictionary. I am not saying that we should get rid of all explicit test settings, but maybe establish a subset of problems (for instance a small network `Flatten -> Linear -> Activation -> Linear -> Loss` and a small CNN) which are then built with and without bias and with different activation functions

f-dangel: Among the line of thoughts above: Maybe something like
```python

def make_simple_cnn_fn(act_cls):
    def make_simple_cnn():
        return torch.nn.Sequential(
            torch.nn.Conv2d(...),
            act_cls(),
            torch.nn.Flatten(),
            torch.nn.Linear(...),
        ),
    
    return make_simple_cnn
```

and loop over different activation functions to create the entries in `SECOND_ORDER_SETTINGS`.

f-dangel: Please use this approach to add such a simple CNN test for the new activations `ELU`, ...

sbharadwajj: Where should this function go? Somewhere central so we can call this function right?
How about `test/networks.py`?

f-dangel: `test/extensions/automated_settings.py`

sbharadwajj:
Changes:
1. typo of `SECONDORDER_SETTINGS`
2. small function to make a simple CNN for the purpose of activation function settings only
3. removed `ELU` and `SELU` because their test fails (fails because of different values)

test/extensions/secondorder/secondorder_settings.py
```diff
@@ -65,11 +67,85 @@
...
+    make_simple_cnn_setting(torch.nn.Sigmoid, bias=False),
+    make_simple_cnn_setting(torch.nn.Tanh, bias=True),
+    make_simple_cnn_setting(torch.nn.Tanh, bias=False),
+    make_simple_cnn_setting(torch.nn.LeakyReLU, bias=True),
+    make_simple_cnn_setting(torch.nn.LeakyReLU, bias=False),
+    make_simple_cnn_setting(torch.nn.LogSigmoid, bias=True),
+    make_simple_cnn_setting(torch.nn.LogSigmoid, bias=False),
+]
+
+
```
f-dangel, line 86: Please rewrite as `for` loop over activation functions and bias

test/automated_test.py
```diff
@@ -91,6 +92,30 @@ def check_values(list1, list2, atol=atol, rtol=rtol):
...
+    dict_setting = {
+        "input_fn": lambda: torch.rand(3, 3, 7, 7),
+        "module_fn": lambda: make_simple_cnn(act_cls, bias),
+        "loss_function_fn": lambda: torch.nn.CrossEntropyLoss(),
+        "target_fn": lambda: classification_targets((3,), 5),
+    }
+
+    return dict_setting
+
+
```
f-dangel, line 118: Please move to `test/extensions/automated_settings.py` and add a module docstring that explains what it does.

test/automated_test.py
```diff
@@ -91,6 +92,30 @@ def check_values(list1, list2, atol=atol, rtol=rtol):
...
+            torch.nn.Flatten(),
+            torch.nn.Linear(72, 5),
+        )
+
+    dict_setting = {
+        "input_fn": lambda: torch.rand(3, 3, 7, 7),
+        "module_fn": lambda: make_simple_cnn(act_cls, bias),
+        "loss_function_fn": lambda: torch.nn.CrossEntropyLoss(),
+        "target_fn": lambda: classification_targets((3,), 5),
+    }
```
f-dangel, line 114: It would be good to label automatically generated settings as such. We can do that with the `'id_prefix="automated-simple-cnn"'`.

test/automated_test.py
sbharadwajj: Ohh, I didnt see it properly. Will do it

test/extensions/secondorder/secondorder_settings.py
```diff
@@ -65,11 +68,78 @@
     },
 ]
 
+###############################################################################
+#                         test setting: Activation Layers                     #
+###############################################################################
+activations = [ReLU, Sigmoid, Tanh, LeakyReLU, LogSigmoid]
+
```
f-dangel, line 75: Please add `for bias in [True, False]` loop

PR #117, opened 2020-11-24
Automate settings

sbharadwajj:
automated settings for `secondorder_settings.py` and `firstorder_settings.py`


test/extensions/automated_settings.py
```diff
@@ -6,7 +6,7 @@
 ###
 
 
-def make_simple_cnn_setting(act_cls, bias):
+def make_simple_act_setting(act_cls, bias):
```
f-dangel, line 9: Is this function still used, or can it be removed?

test/extensions/secondorder/diag_ggn/diaggnn_settings.py
```diff
@@ -7,20 +7,23 @@
...
 SHARED_SETTINGS = SECONDORDER_SETTINGS
+
+
 LOCAL_SETTINGS = []
 
+
 ###############################################################################
 #                         test setting: Activation Layers                     #
 ###############################################################################
 activations = [ELU, SELU]
```
f-dangel, line 23: no `LogSigmoid` here?

test/extensions/automated_settings.py
sbharadwajj: It is in use for the activation settings.

test/extensions/secondorder/diag_ggn/diaggnn_settings.py
sbharadwajj: No, we have it in `secondorder_settings.py`

sbharadwajj: https://github.com/f-dangel/backpack/blob/f362a4c67551f033d5a8be208d8566421bca36ac/test/extensions/secondorder/secondorder_settings.py#L95 ; here

PR #120, opened 2020-12-01
Extension hook

f-dangel:
Extends `with backpack(...)` context by a hook function which is called after all BackPACK extensions have been called on a module. It can be used to compact quantities and free up unneeded buffers, e.g. individual gradients which can become large.

* Add motivation example in ~`all_in_one_example.py`~ as separate use case in doc
* Add tests that extract the `SumGradSquared` and `BatchL2Grad` quantities via hooks into `with backpack(BatchGrad())`


backpack/__init__.py
```diff
@@ -43,6 +44,15 @@ def __init__(self, *exts: BackpropExtension, debug=False):
...
                 The extensions to activate for the backward pass.
+            extension_hook: Callable, optional (default: None)
+                Function with signature ``torch.nn.Module → None``, that will be called
+                on each module after all BackPACK extensions have been computed. It
+                can be used to reduce memory overhead if the goal is to compute
+                transformations of BackPACK quantities. Information can be compacted
+                during a backward pass and obsolete tensors be freed manually.
+
+                If the callable iterates over the module parameters, it may iterate
+                multiple times over some since the hook acts on the modular level.
```
fKunstner, line 55:
```suggestion
            extension_hook: Callable, optional (default: None)
                Function called on each module after all BackPACK extensions have run. 
                Takes a ``torch.nn.Module`` and returns ``None``. 
                
                Can be used to reduce memory overhead if the goal is to compute
                transformations of BackPACK quantities. Information can be compacted
                during a backward pass and obsolete tensors be freed manually (``del``).

                Note: 
                    If the callable iterates over the module parameters, it may iterate
                    multiple times over some since the hook acts on the modular level.
```

fKunstner: I don't think I understand the note 
```
If the callable iterates over the module parameters, it may iterate
multiple times over some since the hook acts on the modular level.
```
Wouldn't the standard backpack hook be called only once per module, and this hook as well? 
If so, it would iterate over the module parameter only once? Unless you meant `model` parameter?

backpack/__init__.py
```diff
@@ -43,6 +44,17 @@ def __init__(self, *exts: BackpropExtension, debug=False):
...
+                Function called on each module after all BackPACK extensions have run. 
+                Takes a ``torch.nn.Module`` and returns ``None``. 
+                
+                Can be used to reduce memory overhead if the goal is to compute
+                transformations of BackPACK quantities. Information can be compacted
+                during a backward pass and obsolete tensors be freed manually (``del``).
+
+                Note: 
+                    If the callable iterates over the module parameters, it may iterate
+                    multiple times over some since the hook acts on the modular level.
```
fKunstner, line 57:
```
If the callable iterates over the module parameters, it may iterate
multiple times over some since the hook acts on the modular level.
```
Not sure I follow. If all backpack hooks (including this one) act on the module level, it will be called once and iterate over the module parameter once? Did you mean `model` or am I missing something?

fKunstner:
For an example, it might be better to write a standalone one and leave the ``all_in_one`` alone. This hook is an advanced feature where the ``all_in_one`` is a first-steps. Here is a draft: [extension_hook_example.txt](https://github.com/f-dangel/backpack/files/5625186/extension_hook_example.txt)

backpack/__init__.py
fKunstner: Depending on what is happening here, ``test/extensions/implementation/hooks.py`` might need some tweaks

f-dangel: I added [tests to demonstrate the subtleties of extension hooks](https://github.com/f-dangel/backpack/blob/0d5f97e5ca93baff371326c28a30ae5df22b287d/test/extensions/test_hooks.py). The origin is that an extension hook will also be called from the backward hook attached to a `nn.Sequential`.

f-dangel:
Agree on moving the feature to the use case section. I think the current version only works because the `diag_h` field is already computed in the backward passes above, and should otherwise run into [this issue](https://github.com/f-dangel/backpack/blob/0d5f97e5ca93baff371326c28a30ae5df22b287d/test/extensions/test_hooks.py#L70-L91).

Will work on it and ping you.

backpack/__init__.py
```diff
@@ -43,6 +44,17 @@ def __init__(self, *exts: BackpropExtension, debug=False):
...
+                Function called on each module after all BackPACK extensions have run.
+                Takes a ``torch.nn.Module`` and returns ``None``.
+
+                Can be used to reduce memory overhead if the goal is to compute
+                transformations of BackPACK quantities. Information can be compacted
+                during a backward pass and obsolete tensors be freed manually (``del``).
+
+                Note:
+                    If the callable iterates over the module parameters, it may iterate
+                    multiple times over some since the hook acts on the modular level.
```
fKunstner, line 57:
```suggestion
                Note:
                    If the callable iterates over the ``module.parameters()``, the same 
                    parameter may be seen multiple times across calls. This happens 
                    if the parameters are part of multiple modules.
                    For example, the parameters of a `torch.nn.Linear` module in 
                    ``model = torch.nn.Sequential(torch.nn.Linear(...))`` are part of both 
                    the ``Linear`` and the ``Sequential``.
```

fKunstner: Great point. Added to the docstring note.

fKunstner:
> I think the current version only works because the `diag_h` field is already computed in the backward passes above, and should otherwise run into [this issue](https://github.com/f-dangel/backpack/blob/0d5f97e5ca93baff371326c28a30ae5df22b287d/test/extensions/test_hooks.py#L70-L91).

Good point. My naive solution to this would be 
```
def hook(module):
    for p in module.parameters():
        if hasattr(p, "bpext_name"): 
            # do stuff 
            delattr(p, "bpext_name")
```
I think it ensures that 
- the loop accesses a parameter only after the BP extension has computed the result 
- and does not visit the same parameter twice as the BP attribute gets clear

f-dangel:
neat worka

> > I think the current version only works because the `diag_h` field is already computed in the backward passes above, and should otherwise run into [this issue](https://github.com/f-dangel/backpack/blob/0d5f97e5ca93baff371326c28a30ae5df22b287d/test/extensions/test_hooks.py#L70-L91).
> 
> Good point. My naive solution to this would be
> 
> ```
> def hook(module):
>     for p in module.parameters():
>         if hasattr(p, "bpext_name"): 
>             # do stuff 
>             delattr(p, "bpext_name")
> ```
> 
> I think it ensures that
> 
>     * the loop accesses a parameter only after the BP extension has computed the result
> 
>     * and does not visit the same parameter twice as the BP attribute gets clear

neat workaround

f-dangel:
@fKunstner Final check, then merge?

fKunstner (approved):

PR #131, opened 2021-01-26
Convnd support for groups in weight_jac_t

sbharadwajj:


test/core/derivatives/convolution_settings.py
```diff
@@ -24,26 +24,65 @@
...
+#         in_channels=4,
+#         out_channels=6,
+#         kernel_size=2,
+#     ),
+#     "input_fn": lambda: torch.rand(size=(3, 4, 7, 7)),
+#     "device": [torch.device("cpu")],  # optional
+#     "seed": 0,  # optional
+#     "id_prefix": "conv-example",  # optional
+# }
+# CONVOLUTION_SETTINGS.append(example)
```
f-dangel, line 48: Is this legacy from the debugging phase? I think it can be removed so there is no git diff

test/core/derivatives/derivatives_test.py
```diff
@@ -210,15 +216,6 @@ def test_weight_jac_mat_prod_should_fail(problem):
...
 
-@pytest.mark.parametrize(
-    "sum_batch", [True, False], ids=["sum_batch=True", "sum_batch=False"]
-)
-@pytest.mark.parametrize("problem", CONVOLUTION_FAIL_PROBLEMS, ids=CONVOLUTION_FAIL_IDS)
-def test_weight_jac_t_mat_prod_should_fail(problem, sum_batch):
-    with pytest.raises(NotImplementedError):
-        test_weight_jac_t_mat_prod(problem, sum_batch)
-
-
```
f-dangel, line 221: I think there were also `ConvTranspose` problems in `CONVOLUTION_FAIL_PROBLEMS`. The coverage report should provide more insights

test/core/derivatives/convolution_settings.py
sbharadwajj: removed

PR #133, opened 2021-02-02
ADD: replace eingroup with einops.rearrange

sbharadwajj:


requirements/test.txt
```diff
@@ -5,3 +5,4 @@ pytest-optional-tests >= 0.1.1
 pytest-cov
 coveralls
 pytorch-memlab
+einops
```
f-dangel, line 8: This should be in `requirements.txt` as it is required by users to run `backpack`; not only to execute the tests

backpack/utils/conv_transpose.py
```diff
@@ -40,15 +32,7 @@ def extract_weight_diagonal(module, input, grad_output, N):
...
-    if N == 1:
-        grad_output_viewed = grad_output
-    elif N == 2:
-        grad_output_viewed = eingroup("v,n,c,h,w->v,n,c,hw", grad_output)
-    elif N == 3:
-        grad_output_viewed = eingroup("v,n,c,d,h,w->v,n,c,dhw", grad_output)
-    else:
-        raise ValueError("{}-dimensional ConvTranspose is not implemented.".format(N))
-
+    grad_output_viewed = rearrange(grad_output, "v n c ... -> v n c (...)")
```
f-dangel, line 35: This looks exactly identical to what `separate_channels_and_pixels` in `utils/conv.py` does, so we could recycle the function.

backpack/core/derivatives/convnd.py
```diff
@@ -194,15 +183,27 @@ def __same_conv_weight_jac_t(self, module, mat, sum_batch):
...
+        if sum_batch:
+            return reduce(
+                grad_weight,
+                "(v n g i o) ... -> v (g o) i ...",
+                "sum",
+                g=G,
+                v=V,
+                n=N,
+                i=C_in // G,
+                o=C_out // G,
```
f-dangel, line 195: Suggestion: Define 
```
dim={"g": G, "v": V, "n": N, "i": C_in // G, "o": C_out // G}
```
then replace these lines by `**dim`. `dim` can be recycled in the `rearrange` call below, too.

f-dangel:
Since `eingroup` is no longer required, it should be removed.

sbharadwajj:
> Since `eingroup` is no longer required, it should be removed.
> 
> @sbharadwajj could you take care of that?

Hii, remove from where? I think I cant find it in any of the `requirements.txt`

f-dangel:
It's in [`backpack.utils.ein.py`](https://github.com/f-dangel/backpack/blob/development/backpack/utils/ein.py).

sbharadwajj:
> It's in [`backpack.utils.ein.py`](https://github.com/f-dangel/backpack/blob/development/backpack/utils/ein.py).

removed

PR #147, opened 2021-04-16
in preparation of support for RNN, reduce output of each module if tuple

tjs111:
The output of an RNN module is tuple (output, h_n) ([RNN](https://pytorch.org/docs/stable/generated/torch.nn.RNN.html))
Similarly the output of LSTM and GRU is (output, ...) ([LSTM](https://pytorch.org/docs/stable/generated/torch.nn.LSTM.html), [GRU](https://pytorch.org/docs/stable/generated/torch.nn.GRU.html))
Therefore, this output has to be reduced in the backward hook to be compatible with backpack calls.

### DIFF
--- a/backpack/__init__.py
+++ b/backpack/__init__.py
@@ -133,11 +133,13 @@ class disable:
 
 def hook_store_io(module, input, output):
     """Saves the input and output as attributes of the module.
-
     Args:
-        module: module
+        module: the module on which to save the params
         input: List of input tensors
-        output: output tensor
+        output: output tensor, result of module(input)
+    Result:
+        list of inputs with index i is saved as module.input[i]
+        output is reduced to single output and saved as module.output
     """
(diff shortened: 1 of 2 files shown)


backpack/__init__.py
```diff
@@ -142,7 +142,11 @@ def hook_store_io(module, input, output):
     if disable.should_store_io() and torch.is_grad_enabled():
         for i in range(len(input)):
             setattr(module, "input{}".format(i), input[i])
-        module.output = output
+        if isinstance(output, tuple):
+            # is true for RNN,GRU,LSTM which return tuple (output, ...)
+            module.output = output[0]
+        else:
+            module.output = output
```
f-dangel, line 149: Cool. I know the docstring of this function has potential for improvement. Could you update it as well?

tjs111: Sure. Is done now.

PR #152, opened 2021-05-03
Working tutorial for custom module

tjs111:
This tutorial shows how to support your own custom module.

### DIFF
--- a/backpack/extensions/backprop_extension.py
+++ b/backpack/extensions/backprop_extension.py
@@ -1,3 +1,4 @@
+"""Implements the backpropagation mechanism."""
 import warnings
 
 from torch.nn import Sequential
@@ -10,8 +11,7 @@ FAIL_SILENT = "SILENT"
 
 
 class BackpropExtension:
-    """
-    Base class for the BackPACK extensions.
+    """Base class for the BackPACK extensions.
 
     Descendants of this class need to
     - define in what field to save results
@@ -30,19 +30,20 @@ class BackpropExtension:
     __external_module_extensions = {}
 
     def __init__(self, savefield, module_exts, fail_mode=FAIL_ERROR):
-        """
-        Parameters
-        ----------
-        savefield : str
-            Where to save results
-        module_exts : dict
-            Dictionary mapping module classes to `ModuleExtension` instances
-        fail_mode : str, optional
-            Behavior when encountering an unknown layer.
-            Can be
-            - "ERROR": raise a NotImplementedError
-            - "WARN": raise a UserWarning
-            - "SILENT": skip the module silently
+        """Initializes parameters.
+
+        Args:
+            savefield(str):
+                Where to save results
+            module_exts(dict):
+                Dictionary mapping module classes to `ModuleExtension` instances
+            fail_mode(str, optional):
+                Behavior when encountering an unknown layer.
+                Can be
+                - "ERROR": raise a NotImplementedError
+                - "WARN": raise a UserWarning
+                - "SILENT": skip the module silently
+                Defaults to FAIL_ERROR = "ERROR"
         """
         self.savefield = savefield
         self.__module_extensions = {
@@ -54,10 +55,19 @@ class BackpropExtension:
 
     @classmethod
     def add_module_extension(cls, module, extension):
+        """Adds an external module mapping.
+
+        Args:
+            module(torch.nn.Module): The module that is supposed to be extended
+            extension(backpack.extensions.module_extension.ModuleExtension):
+                The custom extension of that module.
+        """
         cls.__external_module_extensions[module] = extension
 
     def __get_module_extension(self, module):
         module_extension = self.__module_extensions.get(module.__class__)
+        if module_extension is None:
+            module_extension = self.__external_module_extensions.get(module.__class__)
 
         if module_extension is None:
 
@@ -82,5 +92,12 @@ class BackpropExtension:
         return module_extension.apply
 
     def apply(self, module, g_inp, g_out):
+        """Applies backpropagation.
+
+        Args:
+            module(torch.nn.module): module to perform backpropagation on
+            g_inp(tuple[torch.Tensor]): input gradient
+            g_out(tuple[torch.Tensor]): output gradient
+        """
         module_extension = self.__get_module_extension(module)
         module_extension(self, module, g_inp, g_out)
--- a/docs_src/examples/use_cases/example_custom_module.py
+++ b/docs_src/examples/use_cases/example_custom_module.py
@@ -1,0 +1,167 @@
+"""Custom module example.
+
+This tutorial shows how to support a custom module in a simple fashion.
+"""
+
+# %%
+#
+import torch
+from torch.autograd import Function
+from torch.nn import CrossEntropyLoss
+
+from backpack import backpack, extend
+from backpack.extensions import BatchGrad
+from backpack.extensions.firstorder.base import FirstOrderModuleExtension
+
+# make deterministic
+torch.manual_seed(0)
+
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+
+# %%
+# Define custom function
+class ScaleFunction(Function):
+    """Function that scales the function with weight."""
+
+    @staticmethod
+    def forward(ctx, input, weight):
+        """Forward pass.
+
+        Args:
+            ctx(Any): context object
+            input(torch.Tensor): input
+            weight(torch.Tensor): weight
+
+        Returns:
+            torch.Tensor: input * weight
+        """
+        ctx.save_for_backward(input, weight)
+        output = input * weight
+        return output
+
+    @staticmethod
+    def backward(ctx, grad_output):
+        """Calculates derivatives wrt to input and weight.
+
+        Args:
+            ctx(Any): context object
+            grad_output(torch.Tensor): output gradient
+
+        Returns:
+            tuple[torch.Tensor, torch.Tensor]: grad_input, grad_weight
+        """
+        input, weight = ctx.saved_tensors
+        grad_input = grad_weight = None
+
+        if ctx.needs_input_grad[0]:
+            grad_input = grad_output * weight
+        if ctx.needs_input_grad[1]:
+            grad_weight = grad_output * input
+
+        return grad_input, grad_weight
+
+
+# %%
+# Define custom module
+class ScaleModule(torch.nn.Module):
+    """Defines the module based on ScaleFunction."""
+
+    def __init__(self, input_size=(1,), weight=2.0):
+        """Initializes scale module.
+
+        Args:
+            input_size: size of input is equivalent to output size
+            weight(float): initializes whole weight vector with weight.
+                Defaults to 2.0.
+        """
+        super(ScaleModule, self).__init__()
+        self.input_size = input_size
+
+        self.weight = torch.nn.Parameter(torch.full(input_size, weight))
+
+    def forward(self, input):
+        """Defines forward pass based on ScaleFunction.
+
+        Args:
+            input(torch.Tensor): input
+
+        Returns:
+            torch.Tensor: Result from ScaleFunction
+        """
+        return ScaleFunction.apply(input, self.weight)
+
+
+# %%
+# Create random data
+batch_size = 10
+input_size = 4
+input = torch.randn(batch_size, input_size)
+input.requires_grad = True
+target = torch.randint(0, 2, (batch_size,))
+
+# %%
+# Test custom module
+scaleModule = ScaleModule(input_size=(input_size,))
+lossfunc = CrossEntropyLoss()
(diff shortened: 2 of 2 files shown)


samuelstanton:
@tjs111 Is this example pretty much complete?

tjs111:
> @tjs111 Is this example pretty much complete?

Yes, now it is pretty much complete.
As soon as the pull request is merged, the example will also be available in a more convenient format in the docs.
If you want to use it immediately note that you also need to add the 2 lines of code in backpack/extensions/backprop_extension.py

backpack/extensions/backprop_extension.py
```diff
@@ -30,19 +30,20 @@ class BackpropExtension:
...
-            - "WARN": raise a UserWarning
-            - "SILENT": skip the module silently
+        """Initializes parameters.
+
+        Args:
+            savefield(str):
+                Where to save results
+            module_exts(dict):
+                Dictionary mapping module classes to `ModuleExtension` instances
+            fail_mode(str, optional):
```
f-dangel, line 40: Remove linebreaks between arguments and their descriptions

backpack/extensions/backprop_extension.py
```diff
@@ -54,10 +55,19 @@ def __init__(self, savefield, module_exts, fail_mode=FAIL_ERROR):
...
+            module(Type[torch.nn.Module]): The module that is supposed to be extended
+            extension(backpack.extensions.module_extension.ModuleExtension):
+                The custom extension of that module.
+        """
         cls.__external_module_extensions[module] = extension
 
     def __get_module_extension(self, module):
         module_extension = self.__module_extensions.get(module.__class__)
+        if module_extension is None:
+            module_extension = self.__external_module_extensions.get(module.__class__)
```
f-dangel, line 70: We need to double-check something here. My python knowledge is not solid enough to tell whether, if called from a child class, this method will write to `BackpropExtension.__external_module_extension`, or the child class' attribute.

```python
# writes to BatchGrad.__external_module_extensions or BackpropExtension.__external_module_extensions?
BatchGrad.add_module_extension(MyLinearImplementation1, BatchGradLinear) 
# writes to Grad.__external_module_extensions or BackpropExtension.__external_module_extensions?
# If the second happens, the first call will be overwritten and thus effectively eliminated
Grad.add_module_extension(MyLinearImplementation1, GradLinear)
```
Needs to be debugged before merging. I suspect it has to be refactored because I already experienced the problem for another project.

docs_src/examples/use_cases/example_custom_module.py
```diff
@@ -0,0 +1,168 @@
+"""Custom module example
+=========================================
+
+This tutorial shows how to support a custom module in a simple fashion.
+"""
+
+# %%
+#
```
f-dangel, line 8: Is this dead code?

docs_src/examples/use_cases/example_custom_module.py
```diff
@@ -0,0 +1,168 @@
...
+# This example scales the input by a factor named "weight".
+class ScaleModule(torch.nn.Module):
+    """Defines the module based on ScaleFunction."""
+
+    def __init__(self, input_size=(1,), weight=2.0):
+        """Initializes scale module.
+
+        Args:
+            input_size: size of input is equivalent to output size
```
f-dangel, line 35: missing argument type

f-dangel: missing optional

docs_src/examples/use_cases/example_custom_module.py
```diff
@@ -0,0 +1,168 @@
...
+            weight(float): Initial value for weight. Defaults to 2.0.
```
f-dangel, line 36: missing optional

docs_src/examples/use_cases/example_custom_module.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+# %%
+# Adding custom module to BackPACK
+# -------------------------
+# Define a custom module (https://pytorch.org/docs/stable/notes/extending.html).
+#
+# This example scales the input by a factor named "weight".
+class ScaleModule(torch.nn.Module):
+    """Defines the module based on ScaleFunction."""
```
f-dangel, line 29: `ScaleFunction` does not exist anymore in this version

docs_src/examples/use_cases/example_custom_module.py
```diff
@@ -0,0 +1,168 @@
...
+            input_size: size of input is equivalent to output size
+            weight(float): Initial value for weight. Defaults to 2.0.
+        """
+        super(ScaleModule, self).__init__()
+        self.input_size = input_size
+
+        self.weight = torch.nn.Parameter(torch.tensor(weight))
+
+    def forward(self, input):
+        """Defines forward pass based on ScaleFunction.
```
f-dangel, line 44: `ScaleFunction` does not exist anymore in this version

docs_src/examples/use_cases/example_custom_module.py
```diff
@@ -0,0 +1,168 @@
...
+
+        Args:
+            input(torch.Tensor): input
+
+        Returns:
+            torch.Tensor: Result from ScaleFunction
```
f-dangel, line 50: `ScaleFunction`

docs_src/examples/use_cases/example_custom_module.py
```diff
@@ -0,0 +1,168 @@
...
+            torch.Tensor: individual gradients
+        """
+        # useful available quantities are
+        # output is saved under field output
+        # print("module.output", module.output)
+        # input i is saved under field input[i]
+        # print("module.input0", module.input0)
+        # gradient of output
+        # print("g_out[0]", g_out[0])
+        return (g_out[0] * module.input0).sum(axis=1)
```
f-dangel, line 89: This only works for inputs with 2 dimensions. A more general solution would be
```
# batch_axis = 0
start_feature_dims = 1
return (g_out[0] * module.input0).flatten(start_dim=start_feature_dims).sum(axis=start_feature_dims)
```

docs_src/examples/use_cases/example_custom_module.py
```diff
@@ -0,0 +1,168 @@
...
+BatchGrad.add_module_extension(ScaleModule, ScaleModuleBatchGrad())
+
+
+# %%
+# Testing custom module
+# -------------------------
+# Create some random data and define a function.
+#
+# Note that using "mean" instead of "sum", leads to a within-batch dependency.
+# This alters the individual gradients in autograd (see later) by a factor batch_size.
```
f-dangel, line 102: Maybe better: This scales the vectors backpropagated by PyTorch by ``1 / batch_size``. This scaling peculiarity is also documented here (https://docs.backpack.pt/en/master/extensions.html#backpack.extensions.BatchGrad).

docs_src/examples/use_cases/example_custom_module.py
```diff
@@ -0,0 +1,168 @@
...
+batch_size = 10
+input_size = 4
+input = torch.randn(batch_size, input_size)
+target = torch.randint(0, 2, (batch_size,))
+
+reduction = ["mean", "sum"][1]
+scaleModule = ScaleModule(input_size=(input_size,))
```
f-dangel, line 109: Rename into `my_module` to improve clarity

docs_src/examples/use_cases/example_custom_module.py
```diff
@@ -0,0 +1,168 @@
...
+# %%
+# Adding custom module to BackPACK
+# -------------------------
+# Define a custom module (https://pytorch.org/docs/stable/notes/extending.html).
+#
+# This example scales the input by a factor named "weight".
+class ScaleModule(torch.nn.Module):
+    """Defines the module based on ScaleFunction."""
+
+    def __init__(self, input_size=(1,), weight=2.0):
```
f-dangel, line 31: Is `input_size` needed in the code?

docs_src/examples/use_cases/example_custom_module.py
```diff
@@ -0,0 +1,168 @@
...
+
+loss.backward()
+
+for param in scaleModule.parameters():
+    print("batch gradient", param.grad)
+
+
+# %%
+# Backward with backpack.
+scaleModule_ext = extend(scaleModule)
```
f-dangel, line 125: I think you have to `clone` the module before extending it because the `extend` functions works inplace

docs_src/examples/use_cases/example_custom_module.py
```diff
@@ -0,0 +1,168 @@
...
+scaleModule_ext.zero_grad()
+loss = lossfunc_ext(scaleModule_ext(input), target)
+print("loss", loss)
+
+with backpack(BatchGrad()):
+    loss.backward()
+
+for param in scaleModule_ext.parameters():
+    print("batch gradient", param.grad)
+    print("individual gradients", param.grad_batch)
```
f-dangel, line 136: Maybe don't print the gradients, but their shape? Similar to here (https://docs.backpack.pt/en/master/basic_usage/example_all_in_one.html#first-order-extensions)

docs_src/examples/use_cases/example_custom_module.py
```diff
@@ -0,0 +1,168 @@
...
+for n in range(batch_size):
+    scaleModule.zero_grad()
+    loss = lossfunc(scaleModule(input[n].unsqueeze(0)), target[n].unsqueeze(0))
+    loss.backward()
+    grad_batch_autograd[n] = scaleModule.weight.grad
+
+print("grad_batch_autograd", grad_batch_autograd)
+
+print(
+    "Does autograd and backpack individual gradients match?",
```
f-dangel, line 163: Grammar

docs_src/examples/use_cases/example_custom_module.py
```diff
@@ -0,0 +1,168 @@
...
+
+with backpack(BatchGrad()):
+    loss.backward()
+
+for param in scaleModule_ext.parameters():
+    print("batch gradient", param.grad)
+    print("individual gradients", param.grad_batch)
+
+print(
+    "Does batch gradient match with individual gradients?",
```
f-dangel, line 139: Grammar

backpack/extensions/backprop_extension.py
tjs111: A quick test: A has a class attribute number, and a classmethod set_number. B and C inherit from A.
```
print(A.number, B.number, C.number)  # 0,0,0
A.set_number(1)
print(A.number, B.number, C.number)  # 1,1,1
B.set_number(2)
print(A.number, B.number, C.number)  # 1,2,1
C.set_number(3)
print(A.number, B.number, C.number)  # 1,2,3
A.set_number(0)
print(A.number, B.number, C.number)  # 0,2,3
```
This result is surprising to me. My explanation is that initially the attributes of B and C point to the same attribute as A. But as soon as they are changed they point to a different object.
Thus, as far as our code is concerned, it is fine if the child class adds the module extension. Should we add an error if the method is called from the parent class BackpropExtension?

docs_src/examples/use_cases/example_custom_module.py
tjs111: I think this one is correct. (?)

docs_src/examples/use_cases/example_custom_module.py
tjs111: Reshape and sum instead. Ok?

docs_src/examples/use_cases/example_custom_module.py
tjs111: No. Deleted.

docs_src/examples/use_cases/example_custom_module.py
f-dangel: Ok.

backpack/extensions/backprop_extension.py
f-dangel: Indeed this will cause a bug for someone in the future. I think the easiest, though not most elegant, solution is to make ``__external_module_extension`` independent among instances (i.e. assign it in the constructor) and make ``add_module_extensions`` a regular method of ``BackpropExtension``. What do you think?

tjs111: I would take the following precautions:
1. Against accidently initializing in BackpropExtension instead of an inherited class - check the type of class.
2. Against accidently overwriting an existing mapping - check if the mapping already exists.

I also realized, that the dictionary is added to the regular dictionary in  ``__init__``. Therefore, the check in ``__get_module_extension`` is not necessary.

If you think this solution is not sufficient, I will change the variable to a regular attribute.

tjs111: After discussion, removed ``__external_module_extension``.

f-dangel:
We don't have automated tests (building the example will fail, but I think we won't be notified automatically) to check the interface for defining custom extensions. A basic test is required before acceptance.

Issue #160, opened 2021-06-10
Batch norm individual gradients


schaefertim:
`BatchNorm` is a special module because it mixes samples within the batch. This needs some special treatment.

PyTorch documentation: [BatchNorm1d](https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm1d.html), [BatchNorm2d](https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm2d.html)

If `BatchNorm` is in evaluation mode (`.training=False`), the saved statistics are used. This is independent of the batch. Therefore, individual gradients are well defined.

However, if `BatchNorm` is in training mode (`.training=True`), the batch statistics are used instead. Therefore, individual gradients are not well defined anymore. I suggest multiple possibilities of realizing individual gradients:
1. Use the saved statistics like in evaluation mode.
2. Use `batch_size=1`, i.e. E(x)=x and Var(x)=0.
3. Use complete batch statistics. This is the only case where `.grad_batch.sum(axis=0)==.grad` holds.
4. Use complete batch statistics and approximate individual gradients by neglecting the mixed terms. This leads to separate individual gradients.

I favor the third alternative. If the other approaches do have some merit, it is also possible to implement a switch. For example requesting `module.batch_norm_mode: str` and execute the given mode if provided.

Note: The [current version](https://github.com/f-dangel/backpack/blob/d17408120c30e21bbc7d038fb10b680375940939/backpack/core/derivatives/batchnorm1d.py#L18) does not bother about different modes. I think it exclusively implements the third alternative, but I need some time to verify this.

f-dangel:
More details on this topic can be found [here](https://f-dangel.github.io/about/#filepostsbatchnormalizationorgStructuralimplicationsofbatchnormalization).

f-dangel:
Basic support added in `1.4.0`, see the [changelog](https://github.com/f-dangel/backpack/blob/master/changelog.md#140---2021-11-12).

Closing.

PR #179, opened 2021-06-20
BatchNorm: derivatives and tests

schaefertim:
Surprisingly, the derivatives can all be compared with the PyTorch derivatives. Therefore, I added it to the derivatives test suite.
Check: Is there anything to delete in the old test suite? (I'm not grasping what the [old test suite](https://github.com/f-dangel/backpack/blob/development/test/automated_test.py) is actually testing)

close https://github.com/fKunstner/backpack-discuss/issues/96
close https://github.com/f-dangel/backpack/issues/160


backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,312 @@
...
+        r_mat -= einsum("nc,vc->vnc", x_hat, sum_3)
+        r_mat += (1.0 / N) * einsum("nc,c,vc->vnc", x_hat, sum_24, sum_46)
+
+        r_mat -= einsum("vnc,c->vnc", mat, sum_567)
+        r_mat += (1.0 / N) * einsum("c,vc->vc", sum_567, sum_46).unsqueeze(1).expand(
+            -1, N, -1
+        )
+        r_mat += (3.0 / N) * einsum("nc,vc,c->vnc", x_hat, sum_127, sum_567)
+
+        return einsum("c,vnc->vnc", factor, r_mat)
```
schaefertim, line 296: This is some old code, that I just copied and cleaned a little bit.
Keep it?
Does it need a checker for training mode?
Does it generalize to more dimensions? (It seems to be written for 0d: inputs of shape (N, C). Therefore, it doesn't support BatchNorm1d: inputs of shape (N, C) and inputs of shape (N, C, L).)
Is the tech report out?

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,312 @@
...
+        """
+        return False
+
+    def hessian_is_diagonal(self) -> bool:
+        """Whether hessian is diagonal.
+
+        Returns:
+            whether hessian is diagonal
+        """
+        return False
```
schaefertim, line 53: Never called in tests. (see test coverage)

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,312 @@
...
+    https://chrisyeh96.github.io/2017/08/28/deriving-batchnorm-backprop.html
+    """
+
+    def __init__(self, n_dim: int):
+        """Initialization.
+
+        Args:
+            n_dim: number of dimensions
+        """
+        self.n_dim = n_dim
```
f-dangel, line 32: Rename `n_dim` -> `N` to keep consistent with other `*NdDerivatives`?

backpack/core/derivatives/batchnorm_nd.py
f-dangel: Related to https://github.com/fKunstner/backpack-discuss/issues/62

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,312 @@
...
+        g_inp: Tuple[Tensor],
+        g_out: Tuple[Tensor],
+        mat: Tensor,
+    ) -> Tensor:
+        return self._jac_t_mat_prod(module, g_inp, g_out, mat)
+
+    def _get_normalized_input_and_var(
+        self, module: Union[BatchNorm1d, BatchNorm2d, BatchNorm3d]
+    ) -> Tuple[Tensor, Tensor]:
+        _n_axis: int = self._get_n_axis(module)
```
f-dangel, line 67: Rename into `N`?

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,312 @@
...
+            1: mean[None, :, None],
+            2: mean[None, :, None, None],
+            3: mean[None, :, None, None, None],
+        }[_n_axis]
+        var_expanded: Tensor = {
+            0: var[None, :],
+            1: var[None, :, None],
+            2: var[None, :, None, None],
+            3: var[None, :, None, None, None],
+        }[_n_axis]
```
f-dangel, line 92: Question: Is this to avoid broadcasting?

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,312 @@
...
+    def _get_denominator(module: Union[BatchNorm1d, BatchNorm2d, BatchNorm3d]) -> int:
+        shape_input: Size = module.input0.shape
+        denominator: int = shape_input[0]
+        for i in range(2, len(shape_input)):
+            denominator *= shape_input[i]
+        return denominator
+
+    @staticmethod
+    def _get_n_axis(module: Union[BatchNorm1d, BatchNorm2d, BatchNorm3d]) -> int:
+        return len(module.input0.shape) - 2
```
f-dangel, line 165: Suggestion: Replace `len(x.shape)` with `x.dim()`

backpack/core/derivatives/batchnorm_nd.py
f-dangel: This function is used by the [`HMP`](https://docs.backpack.pt/en/master/extensions.html#backpack.extensions.HMP) extension, and should be kept. It can be generalized to higher-order batch norms at a later stage, probably only in case of a feature request.

This function assumes train mode is on. A check for that would be good.

I PMed you regarding the tech report.

test/core/derivatives/problem.py
```diff
@@ -95,6 +95,12 @@ def set_up(self):
         self.input_shape = self.make_input_shape()
         self.output_shape = self.make_output_shape()
 
+        if not self.module.training:
+            self.module.train(True)
+            for _ in range(3):
+                self.module(torch.randn_like(self.input))
+            self.module.train(False)
+
```
f-dangel, line 103: What does this code do? Set up the statistics moving averages?

test/core/derivatives/derivatives_test.py
```diff
@@ -240,7 +256,11 @@ def test_weight_jac_t_mat_prod(problem, sum_batch, save_memory, V=3):
     problem.tear_down()
 
 
-@pytest.mark.parametrize("problem", PROBLEMS_WITH_WEIGHTS, ids=IDS_WITH_WEIGHTS)
+@pytest.mark.parametrize(
+    "problem",
+    PROBLEMS_WITH_WEIGHTS + BATCH_NORM_PROBLEMS,
+    ids=IDS_WITH_WEIGHTS + BATCH_NORM_IDS,
+)
```
f-dangel, line 263: Needs discussion. My intuition was predicting that this would fail for `sum_batch=False`.

test/core/derivatives/derivatives_test.py
```diff
@@ -217,7 +229,11 @@ def test_weight_hh_l0_jac_t_mat_prod(problem, sum_batch, V=4):
     [True, False],
     ids=["save_memory=True", "save_memory=False"],
 )
-@pytest.mark.parametrize("problem", PROBLEMS_WITH_WEIGHTS, ids=IDS_WITH_WEIGHTS)
+@pytest.mark.parametrize(
+    "problem",
+    PROBLEMS_WITH_WEIGHTS + BATCH_NORM_PROBLEMS,
+    ids=IDS_WITH_WEIGHTS + BATCH_NORM_IDS,
+)
```
f-dangel, line 236: Needs discussion. My intuition was predicting that this would fail for `sum_batch=False`.

f-dangel: Could be related to the initialization of the `weight` parameter.

backpack/core/derivatives/batchnorm_nd.py
schaefertim: Added checker.
Added tech report title.
Moved it to the parent class -> will throw `NotImplementedError` in most cases, but works for 0d

backpack/core/derivatives/batchnorm_nd.py
schaefertim: For consistency, I agree. However I chose `n_dim`, because at the moment we are using `N` for
- `batch_size`
- number of free dimensions
If we have a good name for the second one, it would be better than `N`.
Do you have a suggestion other than `n_dim`?
Should I just change it to `N` for consistency?

backpack/core/derivatives/batchnorm_nd.py
schaefertim: Two reasons:
- Broadcasting doesn't work here (for whatever reason)
- I'm always afraid of broadcasting, because I don't know what happens if different dimensions match in size.

The common underlying issue is my lack of broadcasting knowledge. If you know it better, you can enlighten me.

Separate issue: Probably better do `var=` instead of `var_expanded` (same for mean)?

test/core/derivatives/problem.py
schaefertim: Exactly.

Alternatively, we could set the running stats to some random values. Is this better?

schaefertim: Advantage of the current approach is that the same pattern works for the extension problems.

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,325 @@
...
+        if module.training:
+            denominator: int = self._get_denominator(module)
+            x_hat, var = self._get_normalized_input_and_var(module)
+            ivar = 1.0 / (var + module.eps).sqrt()
+
+            equation = {
+                0: "vni,i->vni",
+                1: "vnil,i->vnil",
+                2: "vnihw,i->vnihw",
+                3: "vnidhw,i->vnidhw",
```
f-dangel, line 113: change `i` to `c`

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,325 @@
...
+        g_out: Tuple[Tensor],
+        mat: Tensor,
+    ) -> Tensor:
+        _n_axis = self._get_n_axis(module)
+        out = mat.unsqueeze(1)
+        for _ in range(_n_axis):
+            out = out.unsqueeze(-1)
+        dim_expand: List[int] = [-1, module.input0.shape[0], -1]
+        for n in range(_n_axis):
+            dim_expand.append(module.input0.shape[2 + n])
```
f-dangel, line 215: Extract into helper function

schaefertim:
Major refactoring to batchnorm_nd.py:
- Removed the child classes BatchNorm1dDerivatives (and 2d, and 3d). They were deprecated, because `N` is retrieved dynamically.
- A lot of functionality is moved to helper functions.

test/core/derivatives/problem.py
f-dangel: There are other modules, like `Dropout`, which we maybe want to test with `train(False)` in the future as they behave differently in evaluation mode. Maybe the `if` condition must be more specific to not run additional forward passes for those modules.

schaefertim: Could also argue the other way around. It is safer to just run some forward passes to be sure that the derivatives stay correct. Thus, it is a trade-off between test execution time and safety.
I argue, that the cost of three additional forward passes is acceptable.

Addtional argument (same as above): For extension test problems, it would be very difficult to write an if-condition to test whether any child contains a BatchNorm module.

However, I could live with writing a specific if-condition in this case and leaving it general for the extensions.

test/core/derivatives/derivatives_test.py
schaefertim: Is this still an open issue? Should we discuss this in person?

f-dangel: Yes. This must probably be discussed in person, unless I find an explanation that makes sense to me.

f-dangel: I think I convinced myself why this works. My intuition was tied too much to the notion of individual gradients, but if you just consider the bare autodiff operations, it makes sense (please confirm):

Consider the mapping `input = [input[0,:], ..., input[N-1,:]] -> output = [output[0,:], ..., output[N-1,:]]` by a parameter `weight`. The mapping can be anything (including batch norm),

* `sum_batch=False` applies the transpose weight Jacobian to every slice in the output. The result is
  ```python
  [
     result[0,:] = (d output[0,:] / d weight)^T mat,
     ...
     result[N-1,:] = (d output[N-1,:] / d weight)^T mat,
  ]
  ```
  My confusion here was that I considered `result[n,:]` the individual gradient for sample `n`, which is not defined
  if `loss[n]` depends on all inputs. But of course such a Jacobian-vector product can be performed since we only care  about derivatives here, rather than the meaningfulness of the resulting operations.
* `sum_batch=True` does basically the same, but returns the sum
  ```python
  (d output[0,:] / d weight)^T mat + ... + (d output[N-1,:] / d weight)^T mat
  ```

f-dangel: Thus I predict that we will run into this problem when testing the `BatchGrad` extension, because there gradients are computed per-sample in a for loop and `BatchNorm([x1, x2, ..., xN]) != [BatchNorm(x1), BatchNorm(x2), ..., BatchNorm(xN)]`

test/core/derivatives/problem.py
f-dangel: In my opinion, the criterion for the derivatives to work, no matter how many forward passes, is that they have the correct IO. This is done by PyTorch's `forward_hook` and should work reliably.

Then, the 'only' reason for additional forward passes is filling the moving averages. I currently favor the solution where this is done in the `SETTINGS` entry `'module_fn'`. Like this,
1. the other tests remain unaffected
2. no additional cost from forward passes
3. no need to write specific if conditions
but the code for specifying a setting becomes longer.

Feedback?

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,287 @@
...
+        if module.track_running_stats is False:
+            raise NotImplementedError("Only implemented for track_running_stats=True")
+
+    def hessian_is_zero(self) -> bool:
+        """Whether hessian is zero.
+
+        Returns:
+            whether hessian is zero
+        """
+        return False
```
f-dangel, line 45: I believe this depends on the mode: With `training=True`, the output is nonlinear in the input (Hessian non-zero), but when `training=False`, then the output is linear in the input, hence the Hessian should be zero.

I just merged testing functionality for `hessian_is_zero` into `development` (#183). You could either merge this into your branch and verify what I wrote by adding a test. Alternatively, you can assume that the old code assumed `training=False` and throw a `NotImplementedError` if `training=True`.

test/core/derivatives/derivatives_test.py
schaefertim: I have difficulties following your calculations...

However, I feel like one important aspect is missing in this argumentation.
If the forward pass is done with the whole batch, all subsequent computations (especially the backward pass) depend on these batch statistics. Therefore, all computed quantities depend on these batch_statistics.

I would even argue, that individual gradients are well defined in this context.

Consequently, I expect the extensions to go just fine, as long as the forward pass is done with the whole batch, even if the backward pass is executed as a for loop.

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,287 @@
...
+
+    def hessian_is_diagonal(self) -> bool:
+        """Whether hessian is diagonal.
+
+        Returns:
+            whether hessian is diagonal
+        """
+        return False
```
f-dangel, line 53: Should also depend on the mode. But we currently don't have tests for this function, it is only called by some second-order extensions, which are only tested by the old test suite which complicates adding new settings.

I vote for raising a `NotImplementedError` when `training=False`.

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,287 @@
...
+    @staticmethod
+    def _get_n_axis(module: Union[BatchNorm1d, BatchNorm2d, BatchNorm3d]) -> int:
+        return module.input0.dim() - 2
+
+    def _unsqueeze_free_axis(
+        self,
+        module: Union[BatchNorm1d, BatchNorm2d, BatchNorm3d],
+        tensor: Tensor,
+        index_batch: int,
+    ) -> Tensor:
```
f-dangel, line 256: Add docstring, mention that this avoids broadcasting.

test/core/derivatives/problem.py
schaefertim: Yes, I agree and will change it accordingly.

However, for future modules keep in mind the following:
I wrote the whole BatchNormDerivatives and tested it successfully. The next day, I realized that the running_stats are trivial and changed them and suddenly all tests failed.
I guess it is fine, because it was also a little bit stupid, and there are very few modules with such a behaviour.

backpack/core/derivatives/batchnorm_nd.py
schaefertim: Since the interface is `hessian_is_zero(self)`, there is no way of determining the training mode. To determine the training mode I need `module.training`.

backpack/core/derivatives/batchnorm_nd.py
schaefertim: Same issue of missing `module` in interface.

backpack/core/derivatives/batchnorm_nd.py
schaefertim: added docstring

backpack/core/derivatives/batchnorm_nd.py
schaefertim: Solutions:
- leave as it is
- change entire interface
- always raise error

test/core/derivatives/derivatives_test.py
f-dangel: I think what you're saying does not conflict with my description:
* The current computation of individual gradients is a for loop (both forward and backward). Hence the `BatchGrad` tests will fail
* If we modify the test to perform the forward pass full-batch, and the backward pass per-sample in a for loop, the test will pass again. **But** I would be careful calling the result an individual gradient. It is a weight Jacobian-vector product for each the individual summand of the mini-batch loss, but every summand depends on **all** samples.

Having this discussion has already helped me a lot to better understand batch norm. Let's continue this in person next week.

test/core/derivatives/problem.py
f-dangel: Just to be sure I understand the problem you had: The problem was that you didn't set the running stats to a value that differs from 0?

backpack/core/derivatives/batchnorm_nd.py
f-dangel: Damn. I also realized this [here](https://github.com/fKunstner/backpack-discuss/issues/110).

The solution will be to modify the interface. I believe leaving it as is is dangerous because we will forget about it. Always raising an error is also not possible because this will make some tests fail.

I think we are kind of about to get stuck with batch norm. Maybe we should take one step back, and improve the interface before coming back. Added to the agenda for Tuesday.

test/core/derivatives/problem.py
schaefertim: Yes. For most modules the initialization is random. For BatchNorm it is: mean=0, var=1, weight=1, bias=0, which is not random at all.

backpack/core/derivatives/batchnorm_nd.py
schaefertim: changed interface

backpack/core/derivatives/batchnorm_nd.py
schaefertim: changed interface

schaefertim:
Those two tests are consistently failing (not visible on github)
```
test_weight_jac_t_mat_prod[dev=cuda-in=(5,7,3,4,2)-BatchNorm3d(7,eps=1e-05,momentum=0.1,affine=True,track_running_stats=True)-save_memory=True-sum_batch=True]
test_weight_jac_t_mat_prod[dev=cuda-in=(5,7,3,4,2)-BatchNorm3d(7,eps=1e-05,momentum=0.1,affine=True,track_running_stats=True)-save_memory=False-sum_batch=True]
```
I will fix different seeds for them.

fully_documented.txt
```diff
@@ -10,6 +10,7 @@ backpack/core/derivatives/permute.py
 backpack/core/derivatives/lstm.py
 backpack/core/derivatives/linear.py
 backpack/core/derivatives/adaptive_avg_pool_nd.py
+backpack/core/derivatives/batchnorm_nd.py
```
f-dangel, line 14: Add BN-related files from tests, too.

test/core/derivatives/batch_norm_settings.py
```diff
@@ -0,0 +1,75 @@
...
+        "module_fn": lambda: _initialize_training_false(
+            torch.nn.BatchNorm2d(num_features=7)
+        ),
+        "input_fn": lambda: torch.rand(size=(5, 7, 3, 4)),
+        "id_prefix": "training=False",
+    },
+    {
+        "module_fn": lambda: torch.nn.BatchNorm3d(num_features=7).train(mode=False),
+        "input_fn": lambda: torch.rand(size=(5, 7, 3, 4, 2)),
+        "id_prefix": "training=False",
```
f-dangel, line 73: Is this case intended to not be initialized?

test/core/derivatives/derivatives_test.py
f-dangel: More in the notes I sent you. Closing here.

f-dangel:
> Those two tests are consistently failing (not visible on github)
> 
> ```
> test_weight_jac_t_mat_prod[dev=cuda-in=(5,7,3,4,2)-BatchNorm3d(7,eps=1e-05,momentum=0.1,affine=True,track_running_stats=True)-save_memory=True-sum_batch=True]
> test_weight_jac_t_mat_prod[dev=cuda-in=(5,7,3,4,2)-BatchNorm3d(7,eps=1e-05,momentum=0.1,affine=True,track_running_stats=True)-save_memory=False-sum_batch=True]
> ```
> 
> I will fix different seeds for them.

I cannot reproduce this on `torch==1.7.1+cu101`:
```bash
pytest -vx test/core/ -k "BatchNorm3d and weight_jac_t and cuda"
```
Output:
```
collected 1795 items / 1787 deselected / 8 selected                                                                                                       

test/core/derivatives/derivatives_test.py::test_weight_jac_t_mat_prod[dev=cuda-in=(5,7,3,4,2)-BatchNorm3d(7,eps=1e-05,momentum=0.1,affine=True,track_running_stats=True)-save_memory=True-sum_batch=True] PASSED [ 12%]
test/core/derivatives/derivatives_test.py::test_weight_jac_t_mat_prod[dev=cuda-in=(5,7,3,4,2)-BatchNorm3d(7,eps=1e-05,momentum=0.1,affine=True,track_running_stats=True)-save_memory=True-sum_batch=False] PASSED [ 25%]
test/core/derivatives/derivatives_test.py::test_weight_jac_t_mat_prod[dev=cuda-in=(5,7,3,4,2)-BatchNorm3d(7,eps=1e-05,momentum=0.1,affine=True,track_running_stats=True)-save_memory=False-sum_batch=True] PASSED [ 37%]
test/core/derivatives/derivatives_test.py::test_weight_jac_t_mat_prod[dev=cuda-in=(5,7,3,4,2)-BatchNorm3d(7,eps=1e-05,momentum=0.1,affine=True,track_running_stats=True)-save_memory=False-sum_batch=False] PASSED [ 50%]
test/core/derivatives/derivatives_test.py::test_weight_jac_t_mat_prod[training=False-dev=cuda-in=(5,7,3,4,2)-BatchNorm3d(7,eps=1e-05,momentum=0.1,affine=True,track_running_stats=True)-save_memory=True-sum_batch=True] PASSED [ 62%]
test/core/derivatives/derivatives_test.py::test_weight_jac_t_mat_prod[training=False-dev=cuda-in=(5,7,3,4,2)-BatchNorm3d(7,eps=1e-05,momentum=0.1,affine=True,track_running_stats=True)-save_memory=True-sum_batch=False] PASSED [ 75%]
test/core/derivatives/derivatives_test.py::test_weight_jac_t_mat_prod[training=False-dev=cuda-in=(5,7,3,4,2)-BatchNorm3d(7,eps=1e-05,momentum=0.1,affine=True,track_running_stats=True)-save_memory=False-sum_batch=True] PASSED [ 87%]
test/core/derivatives/derivatives_test.py::test_weight_jac_t_mat_prod[training=False-dev=cuda-in=(5,7,3,4,2)-BatchNorm3d(7,eps=1e-05,momentum=0.1,affine=True,track_running_stats=True)-save_memory=False-sum_batch=False] PASSED [100%]
```

test/core/derivatives/batch_norm_settings.py
schaefertim: No. I will change it.

schaefertim:
> I cannot reproduce this on `torch==1.7.1+cu101`:

Maybe because I already fixed it to seed=1 in the settings. Try seed=0 to reproduce.

fully_documented.txt
schaefertim: Honestly, there was not much to add.
I additionally added the bn-settings.
I'm leaving out all the bn-extensions because they will be featured in the next pull request.

f-dangel:
Okay. I can reproduce it now with `seed=0`. The results still seem to match well, though not up to the tolerance requested by the test:
```bash
[-0.19422376] versus [-0.1942358]. Ratio of [1.000062]
```
I think setting `seed=1` is better than increasing the tolerance for all tests instead. Agree?

fully_documented.txt
f-dangel: Close.

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,327 @@
...
+        """Multiply with BatchNorm1d residual-matrix.
+
+        Paul Fischer (GitHub: @paulkogni) contributed this code during a research
+        project in winter 2019.
+
+        Details are described in
+
+        `HESSIAN BACKPROPAGATION FOR BATCHNORM`
+        <https://uni-tuebingen.de/en/fakultaeten/mathematisch-naturwissenschaftliche-fakultaet/fachbereiche/informatik/lehrstuehle/methoden-des-maschinellen-lernens/personen/alumni/everyone-else/>
+        <TODO: Wait for tech report upload and add exact link>_
```
f-dangel, line 194: Waiting for the upload. Should be done until Monday.

test/core/derivatives/batch_norm_settings.py
```diff
@@ -0,0 +1,77 @@
...
+        "module_fn": lambda: torch.nn.BatchNorm1d(num_features=7),
+        "input_fn": lambda: torch.rand(size=(5, 7, 4)),
+    },
+    {
+        "module_fn": lambda: torch.nn.BatchNorm2d(num_features=7),
+        "input_fn": lambda: torch.rand(size=(5, 7, 3, 4)),
+    },
+    {
+        "module_fn": lambda: torch.nn.BatchNorm3d(num_features=7),
+        "input_fn": lambda: torch.rand(size=(5, 7, 3, 4, 2)),
```
f-dangel, line 46: Replace `torch.nn.X` by `X`, `torch.rand` by `rand`. Double-check for the entire file.

schaefertim:
> I think setting `seed=1` is better than increasing the tolerance for all tests instead. Agree?

Yes.

test/core/derivatives/batch_norm_settings.py
schaefertim: fixed

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,327 @@
...
+        Args:
+            module: current module to evaluate
+
+        Returns:
+            whether hessian is diagonal
+
+        Raises:
+            NotImplementedError: if module is in evaluation mode
+        """
+        if module.training is True:
```
f-dangel, line 74: Remove `is True` (I will do this)

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,327 @@
...
+        Args:
+            module: current module to evaluate
+
+        Returns:
+            whether hessian is zero
+
+        Raises:
+            NotImplementedError: if module is in evaluation mode
+        """
+        if module.training is True:
```
f-dangel, line 53: Remove `is True` (I will do this)

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,327 @@
...
+            jac_t_mat = denominator * dx_hat
+            jac_t_mat -= dx_hat.sum(
+                self._get_free_axes(module),
+                keepdim=True,
+            ).expand_as(jac_t_mat)
+            equation = "ncl,vmcx,mcx->vncl".replace("l", self.free_index[N]).replace(
+                "x", self.free_index_alternative[N]
+            )
+            jac_t_mat -= einsum(equation, x_hat, dx_hat, x_hat)
+            equation = "vncl,c->vncl".replace("l", self.free_index[N])
```
f-dangel, line 116: Try `vnc...,c->vnc...`

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,327 @@
...
+            jac_t_mat = einsum(equation, jac_t_mat, ivar / denominator)
+            return jac_t_mat
+        else:
+            equation = "c,vncl->vncl".replace("l", self.free_index[N])
```
f-dangel, line 120: Try `c,vnc...->vnc...`

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,327 @@
...
+        mat: Tensor,
+    ) -> Tensor:
+        self._check_parameters(module)
+        N: int = self._get_n_axis(module)
+        if module.training:
+            denominator: int = self._get_denominator(module)
+            x_hat, var = self._get_normalized_input_and_var(module)
+            ivar = 1.0 / (var + module.eps).sqrt()
+
+            equation = "vncl,c->vncl".replace("l", self.free_index[N])
```
f-dangel, line 105: Try `vnc...,c->vnc...`

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,327 @@
...
+            dx_hat: Tensor = einsum(equation, mat, module.weight)
+            jac_t_mat = denominator * dx_hat
+            jac_t_mat -= dx_hat.sum(
+                self._get_free_axes(module),
+                keepdim=True,
+            ).expand_as(jac_t_mat)
+            equation = "ncl,vmcx,mcx->vncl".replace("l", self.free_index[N]).replace(
```
f-dangel, line 112: Try `nc...,vmcx,mcx->vnc...`

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,327 @@
...
+
+    def _weight_jac_mat_prod(
+        self,
+        module: Union[BatchNorm1d, BatchNorm2d, BatchNorm3d],
+        g_inp: Tuple[Tensor],
+        g_out: Tuple[Tensor],
+        mat: Tensor,
+    ) -> Tensor:
+        x_hat, _ = self._get_normalized_input_and_var(module)
+        equation: str = "ncl,vc->vncl".replace(
```
f-dangel, line 135: Try `nc...,vc->vnc...`

backpack/core/derivatives/batchnorm_nd.py
```diff
@@ -0,0 +1,327 @@
...
+        2: "hw",
+        3: "dhw",
+    }
+
+    free_index_alternative: Dict[int, str] = {
+        0: "",
+        1: "x",
+        2: "xy",
+        3: "xyz",
+    }
```
f-dangel, line 327: If the above suggestions work, we should be able to get rid of one dict.

backpack/core/derivatives/batchnorm_nd.py
f-dangel: Document is up: https://uni-tuebingen.de/securedl/sdl-eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE2MjU0ODY0NzAsImV4cCI6MTYyNTU3NjQ2NCwidXNlciI6MCwiZ3JvdXBzIjpbMCwtMV0sImZpbGUiOiJmaWxlYWRtaW5cL1VuaV9UdWViaW5nZW5cL0Zha3VsdGFldGVuXC9NYXROYXRcL0ZhY2hiZXJlaWNoZVwvSW5mb3JtYXRpa1wvTGVocnN0dWVobGVcL01ldGhNYXNjaExlcm5cL0Rva3VtZW50ZVwvVGhlc2VzXC9BYnNjaGx1c3NiZXJpY2h0X0Zpc2NoZXIucGRmIiwicGFnZSI6MTczNDc1fQ.xMe4WmTsyak9J-C7iTxGqTdpYMxMxtGfouyAZgW158I/Abschlussbericht_Fischer.pdf

schaefertim: Link inserted.

backpack/core/derivatives/batchnorm_nd.py
schaefertim: Worked out. I removed both dictionaries since the last one was used only a single time.

PR #180, opened 2021-06-21
[ADD] `SqrtGGNExact` extension

f-dangel:
This PR adds an extension that computes the matrix square root of the generalized Gauss-Newton, using an exact factorization of the loss Hessian. For details, see Equation (3) of [this paper](https://arxiv.org/abs/2106.02624v1).

Auxiliary:
- `pytest`: Enable flag to report explanations of skipped tests
- Use `pytest fixtures` to separate the `set_up` and `tear_down` calls from the test (this should also be done for other tests, see https://github.com/fKunstner/backpack-discuss/issues/97)

### DIFF
--- a/test/extensions/secondorder/sqrt_ggn/test_sqrt_ggn.py
+++ b/test/extensions/secondorder/sqrt_ggn/test_sqrt_ggn.py
@@ -30,7 +30,7 @@ def problem(request, max_num_params: int = 4000) -> ExtensionsTestProblem:
     case.set_up()
 
     num_params = sum(p.numel() for p in case.model.parameters() if p.requires_grad)
-    if num_params < max_num_params:
+    if num_params <= max_num_params:
         yield case
     else:
         skip(f"Model has too many parameters: {num_params} > {max_num_params}")
--- a/test/extensions/secondorder/sqrt_ggn/test_sqrt_ggn.py
+++ b/test/extensions/secondorder/sqrt_ggn/test_sqrt_ggn.py
@@ -1,0 +1,50 @@
+"""Tests BackPACK's ``SqrtGGNExact`` extension."""
+
+from test.automated_test import check_sizes_and_values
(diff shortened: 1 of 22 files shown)


test/extensions/secondorder/sqrt_ggn/test_sqrt_ggn.py
```diff
@@ -0,0 +1,50 @@
...
+def test_sqrt_ggn_exact(problem: ExtensionsTestProblem):
+    """Compare exact GGN from BackPACK's matrix square root with autograd.
+
+    Args:
+        problem: Test case
+    """
+    autograd_res = AutogradExtensions(problem).ggn()
+    backpack_res = BackpackExtensions(problem).ggn()
+
+    check_sizes_and_values(autograd_res, backpack_res)
```
f-dangel: @schaefertim I haven't used `pytest`'s fixtures before. Do you think this makes the code clearer? I think it separates setup and teardown more clearly from the actual test. The filtering functionality can probably be separated from the initialization, and recycled elsewhere in the tests.

schaefertim: Yes, the test `test_sqrt_ggn_exact` looks a lot cleaner and safer.
However, if one needs to create a separate function `problem` for each set of test problems, it probably is ineffective. My pytest knowledge is limited here...

f-dangel: Same here. My feeling is that `problem` could 'request' a `built_problem` that could be shared among many different tests, and only perform the filtering. Will try to read up on it

f-dangel: @schaefertim [Here](https://github.com/f-dangel/backpack/commit/c75794b) is what I was trying to explain above in code.

PR #194, opened 2021-07-02
Use register_full_backward_hook for torch>=1.9.0

schaefertim:
Change `register_backward_hook` to `register_full_backward_hook` for torch>=1.8.0. This is necessary because `register_backward_hook` is deprecated and might be deleted in a future torch release.

Solves https://github.com/fKunstner/backpack-discuss/issues/94

There are significant changes in how the backpropagation quantities are handled.
These changes are necessary because `register_full_backward_hook` changes the behaviour of the forward hook ([reference](https://discuss.pytorch.org/t/register-full-backward-hook-alters-behaviour-of-forward-hook/123466)).
The old approach is to save the backpropagation quantities on the input and retrieve them on the output.
The new approach is the following. There is a separate dictionary saving the quantities under a hash value. The quantities are saved under <s>`grad_inp[0].data_ptr`</s>`input0.data_ptr()` and retrieved under <s>`grad_out[0].data_ptr()`</s>`output.data_ptr()`. ([data_ptr](https://pytorch.org/docs/stable/generated/torch.Tensor.data_ptr.html))
<s>Both approaches are kept and the old approach (torch<1.8.0) is not deprecated yet.</s>
(edit: 2021-07-12) The new hook is used from PyTorch 1.9.0 onwards. For older versions the old register_backward_hook is kept, but the new saving scheme is adapted.
Note (edit: 2021-07-12): `register_full_backward_hook` has the added benefit that the operations are performed in the exact reverse order as the forward hook.

Additional changes:
- add docstrings and type hints for BackPACKs main classes
- all tests: add `requires_grad=True`
  - necessary to ensure the new hook is executed (apparently the old hook was executed no matter what)
  - error message if `requires_grad=False` is not meaningful -> see below
- delete `test/test_simple_resnet.py` since it executed no tests
- delete `test/extensions/test_hooks.py::test_extension_hook_param_before_savefield_exists`
  - I am not sure what this test does
  - test expects `RuntimeError`, but new hook does not raise it (maybe there is even more to delete/deprecate here?)
- (edit: 2021-7-5) delete `test/extensions/test_hooks` since new hook does not visit nodes multiple times

TODOs:
- [x]  throw meaningful error message if failed to fetch backpropagation quantity, <s>e.g. input has `requires_grad = False`</s>
  - [x] BackpropExtension must specify if it expects a quantity, e.g. `def expects_backpropagation_quantity(module: Module) -> bool` (False for first order, True for second order)
  - [x] don't register hooks on containers (e.g. `Sequential`) instead of using `no_op`<s> -> delete utils/hooks.py (only contains `def no_op`)</s>
  - [x] raise meaningful error message if `backpropagation_quantity is None and extension.expects_backpropagation_quantity(module)`
- [x] discuss: hash value `.data_ptr()`: Is this the correct choice?
  - [docs data_ptr](https://pytorch.org/docs/stable/generated/torch.Tensor.data_ptr.html)
  - [torch issue: feature request](https://github.com/pytorch/pytorch/issues/2569)
  - https://github.com/torch/hash (hash function for cpu Tensors)
  - using `id(g_inp[0])` (equivalent to `hash(g_inp[0])`) failed because sometimes ids changed)
- [x] discuss: graph clean of backpropagation quantities (memory overhead)
  - [x] discuss: test that the saved backpropagation quantity graph is clean (old hook)
  - [x] <s>implement same functionality into new hook</s>
  - [x] delete TODO in  `backpack/extensions/module_extension.py::__save_backprop_quantity`
  - [x] (edit: 2021-07-12) clear saved quantities after `with backpack` environment
- [x] make docstring `backpack::__init__` more meaningful (delete TODO)
- [x] (edit: 2021-07-08) properly handle `input.requires_grad = False` (https://github.com/pytorch/pytorch/issues/56380)
  - [x] torch 1.8.x: <s>throw error if `input.requires_grad is False` since the hook won't be called (alternatively,</s> don't use new hook)
  - [x] torch >= 1.9.0: hook will be fired, but need to handle `grad_input[i] is None`
  - [x] <s>set `input.requires_grad=True` in tests only for torch 1.8.x</s>


backpack/extensions/saved_quantities.py
```diff
@@ -0,0 +1,39 @@
...
+
+        Args:
+            key: id of grad_input[0]
+            quantity: tensor to save
+
+        Returns:
+            whether save was successful
+        """
+        if key in self._saved_quantities:
+            return False
```
schaefertim, line 25: Raise a NotImplementedError instead, because this is not covered in tests.
Expected bahaviour is to sum the quantities
Needed for two cases:
- `retain_graph=True`
- ResNets' branching points

@f-dangel Do we need tests with `retain_graph=True`?
I suggest to leave it as a NotImplementedError and that I create an issue.
```suggestion
        if key in self._saved_quantities:
            raise NotImplementedError
```

backpack/__init__.py
```diff
@@ -160,10 +210,16 @@ def hook_run_extensions(module, g_inp, g_out):
...
     See the `post_backward_hook` argument of the `backpack` context manager for details.
+
+    Args:
+        module: current module
+
+    Raises:
+        RuntimeError: if any extension occurred during the backward hook
     """
     try:
         CTX.get_extension_hook()(module)
```
schaefertim, line 225: This Error does not get thrown in tests anymore.
Maybe the whole method can be deleted/deprecated.
@f-dangel Discuss this in person.

backpack/__init__.py
```diff
@@ -195,7 +250,15 @@ def extend(module: torch.nn.Module, debug=False):
...
-        CTX.add_hook_handle(module.register_backward_hook(hook_run_extensions))
+        if TORCH_VERSION_HIGHER_THAN_1_8_0:
+            CTX.add_hook_handle(module.register_full_backward_hook(hook_run_extensions))
+        else:
+            hook_run_extensions_legacy = functools.partial(
+                hook_run_extensions, use_legacy=True
+            )
+            CTX.add_hook_handle(
+                module.register_backward_hook(hook_run_extensions_legacy)
+            )
```
schaefertim, line 261: Not tested anymore. Should there be a test for the old hook?

backpack/extensions/module_extension.py
```diff
@@ -12,98 +23,195 @@ class ModuleExtension:
...
+            module: current module
+            g_inp: input gradients
+            g_out: output gradients
+            use_legacy: whether to use the legacy backward hook.
+                Deprecated since torch version 1.8.0. Default: False.
+
+        Raises:
+            AssertionError: if there is no saved quantity although extension expects one
+        """
+        bpQuantities = self.__get_backproped_quantity(
```
schaefertim, line 95: make variable lower case

backpack/extensions/module_extension.py
```diff
@@ -12,98 +23,195 @@ class ModuleExtension:
...
+            extension: current BackPACK extension
+            reference_tensor: the output Tensor of the current module
+            use_legacy: whether to use the legacy backward hook.
+                Deprecated since torch version 1.8.0. Default: False.
+
+        Returns:
+            the backpropagation quantity
+        """
+        if extension.expects_backpropagation_quantities() is False:
+            return None
```
schaefertim, line 144: delete

schaefertim (commented):
Some suggestions...

backpack/__init__.py
schaefertim: Is tested with new CI.

backpack/extensions/saved_quantities.py
schaefertim: The issue is not precise any more. It is more about the clean-up after a backward hook. A wrote a corresponding TODO in the code.

backpack/__init__.py
f-dangel: This was intended to provide an improved error message and can be removed. The method however cannot be removed.

backpack/__init__.py
```diff
@@ -160,10 +203,16 @@ def hook_run_extensions(module, g_inp, g_out):
         memory_cleanup(module)
 
 
-def run_extension_hook(module):
+def run_extension_hook(module: Module) -> None:
     """Execute the post extensions hook on a module after all BackPACK extensions.
 
     See the `post_backward_hook` argument of the `backpack` context manager for details.
```
f-dangel, line 209: post_extension_hook

docs_src/examples/use_cases/example_trace_estimation.py
```diff
@@ -225,7 +225,7 @@ def hutchinson_trace_autodiff_blockwise(V):
 
     plt.semilogx(
         V_list,
-        trace_estimates,
+        [trace_estimate.to("cpu") for trace_estimate in trace_estimates],
```
f-dangel, line 228: Suggestion: `trace_estimate.cpu()`

docs_src/examples/use_cases/example_first_order_resnet.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, C_in=1, C_hid=5, input_dim=(28, 28), output_dim=10):
     def forward(self, x):
         residual = self.shortcut(x)
         x = self.conv2(F.relu(self.conv1(x)))
-        x += residual
+        x = x + residual  # don't use: x += residual
```
f-dangel, line 56: will the inplace operation cause a problem?

test/extensions/problem.py
```diff
@@ -145,10 +145,10 @@ def forward_pass(self, sample_idx=None):
                 input, output, loss, each with batch axis first
         """
         if sample_idx is None:
-            input = self.input.clone().detach()
+            input = self.input.clone()
             target = self.target.clone().detach()
         else:
-            target = self.target.split(1, dim=0)[sample_idx].detach()
+            target = self.target.split(1, dim=0)[sample_idx]
             input = self.input.split(1, dim=0)[sample_idx].detach()
```
f-dangel, line 152: Try removing the other `detach` ops, too.

test/extensions/test_hooks.py
```diff
@@ -1,91 +0,0 @@
...
-        for param_id, visits in params_visited.items():
-            if visits == 0:
-                raise ValueError(f"Hook never visited param {param_id}")
-            elif visits == 1:
-                pass
-            else:
-                raise AssertionError(f"Hook visited param {param_id} {visits} times ")
-
-    with pytest.raises(AssertionError):
-        check()
```
f-dangel, line 67: This test should still work under any PyTorch version. Try reintroducing.

f-dangel: **Update:** This test was required because the old approach installed hooks on containers, like `Sequential`, which it does not anymore after this PR.

Still, this means that the tested behavior could happen with user-defined containers, as they will have hooks installed (only `Sequential` and `Reduce` won't have hook). How to proceed? Test that each parameter is visited once?

backpack/extensions/firstorder/batch_grad/__init__.py
```diff
@@ -89,3 +89,6 @@ def get_subsampling(self) -> Union[List[int], None]:
             means all samples will be considered.
         """
         return self._subsampling
+
+    def expects_backpropagation_quantities(self) -> bool:  # noqa: D102
+        return False
```
f-dangel, line 94: Can this be moved to a `FirstOrderBackpropExtension` base class that inherits from `BackpropExtension`? Same for the other first-order extensions

backpack/extensions/secondorder/hbp/__init__.py
```diff
@@ -69,6 +69,9 @@ def get_backprop_strategy(self):
     def get_ea_strategy(self):
         return self.ea_strategy
 
+    def expects_backpropagation_quantities(self) -> bool:  # noq: D102
+        return True
+
```
f-dangel, line 74: Can this be moved to a `SecondOrderBackpropExtension` parent class that inherits from `BackpropExtension`? Same for the other second-order extensions.

docs_src/examples/use_cases/example_first_order_resnet.py
schaefertim: From torch 1.9.0 onwards it throws the following error
```
RuntimeError: Output 0 of BackwardHookFunctionBackward is a view and is being modified inplace. This view was created inside a custom Function (or because an input was returned as-is) and the autograd logic to handle view+inplace would override the custom backward associated with the custom Function, leading to incorrect gradients. This behavior is forbidden. You can remove this warning by cloning the output of the custom Function.
```
Seems like using a backward hook in combination with inplace and views does not work.

backpack/__init__.py
schaefertim: renamed

docs_src/examples/use_cases/example_trace_estimation.py
schaefertim: Yes.

test/extensions/problem.py
schaefertim: Worked.

backpack/extensions/firstorder/batch_grad/__init__.py
schaefertim: Created a base class.

backpack/extensions/secondorder/hbp/__init__.py
schaefertim: Created a base class.

test/extensions/test_hooks.py
schaefertim: I could
- reintroduce the test
- change the setup to such a custom user module
- throw an error message such that the user gets an explanation about containers (this point is difficult, because with ResNets a node will indeed be visited more than once without throwing an error)

Alternatively, I could just continue without this test.

backpack/utils/__init__.py
```diff
@@ -6,3 +6,5 @@
 VERSION_1_9_0 = packaging.version.parse("1.9.0")
 VERSION_1_8_0 = packaging.version.parse("1.8.0")
 VERSION_1_6_0 = packaging.version.parse("1.6.0")
+TORCH_VERSION_HIGHER_THAN_1_8_0 = VERSION_1_8_0 <= TORCH_VERSION
+TORCH_VERSION_HIGHER_THAN_1_9_0 = VERSION_1_9_0 <= TORCH_VERSION
```
f-dangel, line 10: Flip operands to improve readability

backpack/utils/hooks.py
```diff
@@ -1,6 +1,25 @@
...
+
+
+def no_op_apply(
+    extension: BackpropExtension,
+    module: Module,
+    grad_inp: Tuple[Tensor],
+    grad_out: Tuple[Tensor],
+    use_legacy: bool = False,
+) -> None:
+    pass
```
f-dangel, line 25: Needs documentation, especially `use_legacy`. Add file to `fully_documented.txt`.

test/extensions/test_hooks.py
f-dangel: I don't fully understand your points -> personal discussion

backpack/extensions/firstorder/base.py
```diff
@@ -1,6 +1,14 @@
+from backpack.extensions.backprop_extension import BackpropExtension
 from backpack.extensions.module_extension import ModuleExtension
 
 
 class FirstOrderModuleExtension(ModuleExtension):
     def backpropagate(self, ext, module, g_inp, g_out, bpQuantities):
         return None
```
f-dangel, line 7: Add file to fully-documented. Replace `return None` by `pass` (cleaner for 'does nothing')

backpack/__init__.py
```diff
@@ -1,44 +1,49 @@
...
-            all BackPACK extensions have run. Takes a ``torch.nn.Module`` and returns
-            ``None``. Default: ``None`` (no operation will be formed).
+        Enables the BackPACK extensions passed as arguments in the
+        :code:`backward` calls inside the current :code:`with` block.
+
+        Args:
+            exts: Extensions to activate in the backward pass.
+            extension_hook: Function called on each module after
+                all BackPACK extensions have run. Takes a ``torch.nn.Module`` and returns
+                ``None``. Default: ``None`` (no operation will be formed).
```
f-dangel, line 36: `formed` -> `performed`

backpack/__init__.py
```diff
@@ -1,44 +1,49 @@
...
             during a backward pass and obsolete tensors be freed manually (``del``).
 
-            .. note::
-
-                If the callable iterates over the ``module.parameters()``, the same
-                parameter may be seen multiple times across calls. This happens
-                if the parameters are part of multiple modules.
-                For example, the parameters of a `torch.nn.Linear` module in
-                ``model = torch.nn.Sequential(torch.nn.Linear(...))`` are part of
-                both the ``Linear`` and the ``Sequential``.
```
f-dangel, line 36: Needs confirmation by a test: Extension hook visits every parameter once and is not called on containers which would lead to visiting the contained parameters more than once.

backpack/__init__.py
```diff
@@ -66,6 +74,15 @@ def __enter__(self):
         CTX.set_extension_hook(self.extension_hook)
 
     def __exit__(self, type, value, traceback):
+        """Leave backpack environment.
+
+        Args:
+            type: .
+            value: .
+            traceback: .
+        """
```
f-dangel, line 83: Re types and argument descriptions: I found [this answer](https://stackoverflow.com/a/58808179) and [this example code](https://github.com/python/typeshed/blob/095464874accef1cc5303f325ccabe88756bcd8b/stdlib/2and3/contextlib.pyi#L67-L69).

backpack/__init__.py
```diff
@@ -91,33 +108,45 @@ class disable:
...
 
     def __exit__(self, type, value, traceback):
-        """Set input/output storing to old value."""
+        """Set input/output storing to old value.
+
+        Args:
+            type: .
+            value: .
+            traceback: .
+        """
```
f-dangel, line 125: Add types and arg descriptions

backpack/__init__.py
```diff
@@ -91,33 +108,45 @@ class disable:
...
         disable.store_io = self.old_store_io
 
     @staticmethod
-    def should_store_io():
-        """Return whether input and output should be stored."""
+    def should_store_io() -> bool:
+        """Return whether input and output should be stored.
```
f-dangel, line 130: Suggestion: during a forward pass

backpack/__init__.py
```diff
@@ -91,33 +108,45 @@ class disable:
...
     """Saves the input and output as attributes of the module.
 
     The list of inputs with index i is saved as module.input[i]
     The output is reduced to single output tensor and saved as module.output
 
     Args:
-        module (torch.nn.Module): the module on which to save the params
-        input (list): List of input tensors
-        output (torch.Tensor or tuple): result of module(input)
+        module: the module on which to save the params
```
f-dangel, line 147: Replace 'params' with 'inputs/outputs'

backpack/__init__.py
f-dangel: I still see the old name. Reopening

backpack/__init__.py
```diff
@@ -142,13 +174,24 @@ def memory_cleanup(module):
...
+        g_inp: input gradients
+        g_out: output gradients
+    """
     for backpack_extension in CTX.get_active_exts():
         if CTX.get_debug():
             print("[DEBUG] Running extension", backpack_extension, "on", module)
         backpack_extension.apply(module, g_inp, g_out)
 
-    run_extension_hook(module)
+    post_extension_hook(module)
```
f-dangel, line 194: [Ignore for now] Need more documentation which modules have a backward hook installed that will lead to execution of  `post_extension_hook`.

backpack/__init__.py
```diff
@@ -160,10 +203,16 @@ def hook_run_extensions(module, g_inp, g_out):
...
+def post_extension_hook(module: Module) -> None:
     """Execute the post extensions hook on a module after all BackPACK extensions.
 
     See the `post_backward_hook` argument of the `backpack` context manager for details.
+
+    Args:
+        module: current module
+
+    Raises:
+        RuntimeError: if any error occurred during the backward hook
```
f-dangel, line 215: I think there is no need anymore to specifically raise an error in the body. The traceback will indicate that an error occured in `post_extension_hook`. This should be enough. Remove the `try ... except`. We could even inline the remaining call `CTX.get_extension_hook()(module)` in `hook_run_extensions`

backpack/__init__.py
```diff
@@ -172,19 +221,18 @@ def run_extension_hook(module):
         raise RuntimeError(f"Post extensions hook failed: {message}")
 
 
-def extend(module: torch.nn.Module, debug=False):
+def extend(module: Module, debug: bool = False) -> Module:
     """Extends a ``module`` to make it BackPACK-ready.
 
     If the ``module`` has children, e.g. for a ``torch.nn.Sequential``,
     they will also be extended.
```
f-dangel, line 228: Needs more documentation: Which modules get hooks installed. How are container modules treated?

backpack/__init__.py
```diff
@@ -195,7 +243,17 @@ def extend(module: torch.nn.Module, debug=False):
     module_was_already_extended = getattr(module, "_backpack_extend", False)
     if not module_was_already_extended:
         CTX.add_hook_handle(module.register_forward_hook(hook_store_io))
-        CTX.add_hook_handle(module.register_backward_hook(hook_run_extensions))
+        _register_hooks(module)
```
f-dangel, line 246: [No action required for now] Discuss that there is no need to install hooks on container modules such as `Sequential`

backpack/__init__.py
```diff
@@ -195,7 +243,17 @@ def extend(module: torch.nn.Module, debug=False):
...
         module._backpack_extend = True
 
     return module
+
+
+def _register_hooks(module: Module) -> None:
+    if is_no_op(module):
+        return
```
f-dangel, line 254: [No action required for now] Discuss where it should be decided that a module will get hooks installed.

backpack/utils/hooks.py
schaefertim: `use_legacy` was expendable after all. Added documentation.

backpack/__init__.py
schaefertim: Awesome!

backpack/__init__.py
schaefertim: Indeed. Now the method is deleted.

backpack/__init__.py
schaefertim: I deleted the method and put it inline.

f-dangel:
# Thoughts on BackPACK's hook policy

BackPACK installs forward module hooks that save I/O and backward hooks that run extensions, post extension hooks, and then delete the I/O from the forward pass.

**Current policy:** The hooks are added by the [extend](https://github.com/f-dangel/backpack/blob/2c632c9fd2e54119595fc32745eaec4d3babdb4b/backpack/__init__.py#L175-L201) function. Currently, the criterion for a module to get hooks installed is that it [has not already been extended](https://github.com/f-dangel/backpack/blob/2c632c9fd2e54119595fc32745eaec4d3babdb4b/backpack/__init__.py#L196). There are some downsides of this approach:

1.  There exist container modules that do not require additional code be executed for BackPACK's quantities. When a hook encounters such a container, it [returns a no<sub>op</sub>](https://github.com/f-dangel/backpack/blob/2c632c9fd2e54119595fc32745eaec4d3babdb4b/backpack/extensions/backprop_extension.py#L82-L83). This no<sub>op</sub> is necessary in order to reach the code that frees up the I/O saved by the container module during a forward pass.
2.  BackPACK's [post<sub>extension</sub><sub>hook</sub>](https://github.com/f-dangel/backpack/blob/2c632c9fd2e54119595fc32745eaec4d3babdb4b/backpack/__init__.py#L151) is executed in a backward hook. Sometimes it is executed on the container, and on each of the modules it contains. Problems:
    -   **Multiple traversals:** If the post<sub>extension</sub><sub>hook</sub> iterates over `module.parameters()`, it will traverse some parameters multiple times. This is confusing, but [documented](https://github.com/f-dangel/backpack/blob/2c632c9fd2e54119595fc32745eaec4d3babdb4b/backpack/__init__.py#L29-L36) and [tested](https://github.com/f-dangel/backpack/blob/2c632c9fd2e54119595fc32745eaec4d3babdb4b/test/extensions/test_hooks.py#L41-L67).
    -   **Early execution:** It can happen that the post<sub>extension</sub><sub>hook</sub> is first executed on the container at a time where BackPACK's quantities have not yet been computed by the submodules. If the post<sub>extension</sub><sub>hook</sub> tries to access BackPACK quantity savefields, the quantity won't be there. This is tested [here](https://github.com/f-dangel/backpack/blob/2c632c9fd2e54119595fc32745eaec4d3babdb4b/test/extensions/test_hooks.py#L70-L91).

**Proposed change:** Do not install any hooks on container modules. Pros:

1.  Eliminates the need to return a no<sub>op</sub> in the first downside of the old approach. Less I/O saved and hooks installed. Might therefore improve memory performance.
2.  Resolves the problem **multiple traversal** and **early execution** for the post<sub>extension</sub><sub>hook</sub>, since containers won't have any hook anymore.

Cons:

1.  No debug print messages for containers from [extending](https://github.com/f-dangel/backpack/blob/2c632c9fd2e54119595fc32745eaec4d3babdb4b/backpack/__init__.py#L189-L190) and [executing extensions](https://github.com/f-dangel/backpack/blob/2c632c9fd2e54119595fc32745eaec4d3babdb4b/backpack/__init__.py#L148).

backpack/context.py
```diff
@@ -1,56 +1,104 @@
...
-    def set_active_exts(active_exts):
+    def set_active_exts(active_exts: Iterable[BackpropExtension]) -> None:
+        """Set the active backpack extensions.
+
+        Args:
+            active_exts: the extensions
+        """
         CTX.active_exts = tuple()
         for act_ext in active_exts:
             CTX.active_exts += (act_ext,)
```
f-dangel, line 28: Suggestion: `CTX.active_exts = tuple(active_exts)`

backpack/context.py
```diff
@@ -1,56 +1,104 @@
...
+        Args:
+            *extension_classes: classes to test
+
+        Returns:
+            whether at least one of the specified extensions is active
+        """
         for backpack_ext in CTX.get_active_exts():
             if isinstance(backpack_ext, extension_classes):
                 return True
         return False
```
f-dangel, line 68: Suggestion: `return any(isinstance(ext, extension_classes) for ext in CTX.get_active_exts())`

backpack/context.py
```diff
@@ -1,56 +1,104 @@
...
+        """Whether debug mode is active.
+
+        Returns:
+            whether debug mode is active
+        """
         return CTX.debug
 
     @staticmethod
-    def set_debug(debug):
+    def set_debug(debug) -> None:
```
f-dangel, line 80: `debug: bool`

backpack/extensions/backprop_extension.py
```diff
@@ -74,38 +89,56 @@ def set_module_extension(
...
-                warnings.warn(
-                    "Extension saving to {} ".format(self.savefield)
-                    + "does not have an extension for "
-                    + "Module {}".format(module.__class__)
-                )
+                for _ in module.parameters():
+                    warnings.warn(
+                        f"Extension saving to {self.savefield} does not have an "
+                        f"extension for Module {module.__class__} "
+                        f"although the module has parameters"
```
f-dangel, line 109: This only warns if a module has parameters, although parameter-free modules that don't have an extension should also trigger such a warning.

f-dangel: Try recycling the message string.

backpack/extensions/backprop_extension.py
```diff
@@ -74,38 +89,56 @@ def set_module_extension(
...
         module_extension(self, module, g_inp, g_out)
+
+    @abc.abstractmethod
+    def expects_backpropagation_quantities(self) -> bool:
+        """Whether the extension uses additional backpropagation quantities.
+
+        Returns:
+            Whether the extension uses additional backpropagation quantities.
+        """
+        return
```
f-dangel, line 140: `pass` instead?

backpack/extensions/backprop_extension.py
```diff
@@ -74,38 +89,56 @@ def set_module_extension(
...
+                    )
+                    break
 
-            return no_op
+            return no_op_apply
 
         return module_extension.apply
 
-    def apply(self, module, g_inp, g_out):
+    def apply(
```
f-dangel, line 117: Let's use python's built-in `__call__` method to make an object callable: Rename into `__call__`. Replace `backprop_extension.apply(...)` by `backprop_extension(...)` in  `__init__.py`.

backpack/extensions/module_extension.py
```diff
@@ -12,98 +24,156 @@ class ModuleExtension:
...
         Returns
-        -------
-        bpQuantities:
             Quantities backpropagated w.r.t. the input
         """
         warnings.warn("Backpropagate has not been overwritten")
 
-    def apply(self, ext, module, g_inp, g_out):
-        """
+    def apply(
```
f-dangel, line 72: Let's use python's built-in `__call__` method to make an object callable: Rename into `__call__`. Replace `module_extension.apply(...)` by `module_extension(...)`.

backpack/extensions/module_extension.py
```diff
@@ -12,98 +24,156 @@ class ModuleExtension:
...
-            method.
-            The result of this method will be saved in the savefield of `m.p`.
+    def __init__(self, params: List[str] = None):
+        """Initialization.
+
+        Args:
+            params: List of module parameters that need special treatment.
+                for each param `p` in the list, instances of the extended module `m`
+                need to have a field `m.p` and the class extending `ModuleExtension`
+                need to provide a method with the same signature as the `backprop`
```
f-dangel, line 34: `backpropagate` method

backpack/extensions/module_extension.py
```diff
@@ -12,98 +24,156 @@ class ModuleExtension:
...
-            need to have a field `m.p` and the class extending `ModuleExtension`
-            need to provide a method with the same signature as the `backprop`
-            method.
-            The result of this method will be saved in the savefield of `m.p`.
+    def __init__(self, params: List[str] = None):
+        """Initialization.
+
+        Args:
+            params: List of module parameters that need special treatment.
+                for each param `p` in the list, instances of the extended module `m`
```
f-dangel, line 32: For

backpack/extensions/module_extension.py
```diff
@@ -12,98 +24,156 @@ class ModuleExtension:
...
+                method.
+                The result of this method will be saved in the savefield of `m.p`.
+
+        Raises:
+            NotImplementedError: if child class doesn't have a method for each parameter
         """
         if params is None:
             params = []
 
         self.__params = params
```
f-dangel, line 44: `self.__params = [] if params is None else params`

backpack/extensions/module_extension.py
```diff
@@ -12,98 +24,156 @@ class ModuleExtension:
...
 
         for param in self.__params:
-            extFunc = getattr(self, param, None)
-            if extFunc is None:
+            if hasattr(self, param) is False:
                 raise NotImplementedError
```
f-dangel, line 48: Provide error message

backpack/extensions/module_extension.py
```diff
@@ -12,98 +24,156 @@ class ModuleExtension:
...
+            module: current module
+            g_inp: input gradients
+            g_out: output gradients
+
+        Raises:
+            AssertionError: if there is no saved quantity although extension expects one
+        """
+        bp_quantity = self.__get_backproped_quantity(extension, module.output)
+        if (
+            extension.expects_backpropagation_quantities() is True
```
f-dangel, line 95: Delete `is True`

backpack/extensions/module_extension.py
```diff
@@ -12,98 +24,156 @@ class ModuleExtension:
...
+            and bp_quantity is None
+            and is_loss(module) is False
```
f-dangel, line 97: `and not is_loss`

backpack/extensions/module_extension.py
```diff
@@ -12,98 +24,156 @@ class ModuleExtension:
...
+
+        if extension.expects_backpropagation_quantities():
+            bp_quantity = self.backpropagate(
+                extension, module, g_inp, g_out, bp_quantity
+            )
+            self.__save_backprop_quantity(
+                extension,
+                module.input0,
+                bp_quantity,
+            )
```
f-dangel, line 119: can this be one line?

backpack/extensions/module_extension.py
```diff
@@ -12,98 +24,156 @@ class ModuleExtension:
...
 
-        bpQuantities = self.backpropagate(ext, module, g_inp, g_out, bpQuantities)
+    @staticmethod
+    def __get_backproped_quantity(
+        extension: BackpropExtension, reference_tensor: Tensor
+    ) -> Tensor or None:
+        """Fetch backpropagated quantities attached to the module output.
```
f-dangel, line 125: Add documentation which property of `reference_tensor` is used as identifier

backpack/extensions/module_extension.py
```diff
@@ -12,98 +24,156 @@ class ModuleExtension:
...
     @staticmethod
-    def __backprop_quantities(ext, inp, out, bpQuantities):
-        """Propagate back additional information by attaching it to the module input."""
-
-        setattr(inp, ext.savefield, bpQuantities)
+    def __save_backprop_quantity(
+        extension: BackpropExtension,
+        reference_tensor: Tensor,
+        bpQuantities: Any,
+    ) -> None:
```
f-dangel, line 141: Can this be 3 lines?

backpack/extensions/saved_quantities.py
```diff
@@ -0,0 +1,46 @@
...
+        """Saves the quantity under the specified key.
+
+        Args:
+            key: id of grad_input[0]
+            quantity: tensor to save
+
+        Raises:
+            NotImplementedError: if the key already exists
+        """
+        if key in self._saved_quantities:
```
f-dangel, line 24: `if key in self._saved_quantities.keys()`?

backpack/utils/hooks.py
```diff
@@ -1,6 +1,39 @@
...
+
+    It is used when no hooks are installed (e.g. containers).
+
+    Args:
+        extension: backpack extension
+        module: module without hooks
+        grad_inp: input gradients
+        grad_out: output gradients
+    """
+    pass
```
f-dangel, line 39: Maybe not necessary anymore (see comment on hook policy)

backpack/extensions/module_extension.py
```diff
@@ -12,98 +24,156 @@ class ModuleExtension:
...
+            bpQuantities: backpropagation quantities that should be saved
+        """
+        extension.saved_quantities.save_quantity(
+            reference_tensor.data_ptr(), bpQuantities
+        )
 
-        is_a_leaf = out.grad_fn is None
-        retain_grad_is_on = getattr(out, "retains_grad", False)
-        inp_is_out = id(inp) == id(out)
-        should_retain_grad = is_a_leaf or retain_grad_is_on or inp_is_out
```
f-dangel, line 96: Needs to be introduces. Discuss in person.

backpack/extensions/backprop_extension.py
schaefertim: I would disagree, that parameter-free modules should trigger a warning.
All first order extensions operate in warning mode. It is pretty annoying if they throw a warning for every module that has no extension. 
Note: In the old version, there was never a warning because, the mode was misspelled. That is the reason for the dictionary check of the mode.

If you think a warning is the best approach, I will change it.

backpack/extensions/backprop_extension.py
schaefertim: I'm not sure. If I use `pass`, darglint complains about an excess `Returns` in the docstring. I would keep the `return`.

backpack/extensions/backprop_extension.py
schaefertim: Nice idea. Implemented it.

backpack/extensions/module_extension.py
schaefertim: Nice idea. Implemented it.

backpack/utils/hooks.py
schaefertim: Yes. Deleted it.

backpack/extensions/module_extension.py
schaefertim: Wrote an error message. Please check whether it contains the correct information.

backpack/extensions/saved_quantities.py
schaefertim: I thought so, too. But [this](https://stackoverflow.com/questions/44035287/check-key-exist-in-python-dict/44035382) seems to be the correct way.

backpack/extensions/backprop_extension.py
f-dangel: Okay.

backpack/extensions/saved_quantities.py
f-dangel: Okay.

backpack/extensions/backprop_extension.py
f-dangel: Your explanation makes sense, I approve your version. Good catch with the misspelled `FAIL_MODE`. Could you replace the strings passed to `fail_mode` by all extensions by variables `FAIL_WARN, FAIL_ERROR`?

schaefertim: I placed it into the `FirstOrderBackpropExtension`.

f-dangel: I think it's better to keep `fail_mode` as argument in BackPACK's extensions that are exposed in the API. Your current change breaks the API for users that modified `fail_mode` in their code

schaefertim: Added `fail_mode=FAIL_WARN` to `FirstOrderBackpropExtension`'s arguments.

backpack/extensions/module_extension.py
f-dangel: Confirm.

backpack/extensions/backprop_extension.py
f-dangel: I meant that it should be part of the constructors of `BatchGrad`, etc. (maybe I also don't see the latest version).

schaefertim: This would be a new feature since the [current version on development](https://github.com/f-dangel/backpack/blob/2c632c9fd2e54119595fc32745eaec4d3babdb4b/backpack/extensions/firstorder/batch_grad/__init__.py#L60) does not have a `fail_mode` parameter.

test/extensions/test_hooks.py
f-dangel: Implement the discussed tests, then this will be resolved.

backpack/extensions/module_extension.py
```diff
@@ -12,98 +24,156 @@ class ModuleExtension:
...
 
         for param in self.__params:
             if self.__param_exists_and_requires_grad(module, param):
                 extFunc = getattr(self, param)
-                extValue = extFunc(ext, module, g_inp, g_out, bpQuantities)
-                self.__save(extValue, ext, module, param)
+                extValue = extFunc(extension, module, g_inp, g_out, bp_quantity)
+                self.__save_value_on_parameter(extValue, extension, module, param)
+
+        if extension.expects_backpropagation_quantities():
```
f-dangel, line 113: Add a function `should_backpropagate(extension, module)`, which checks that backpropagation must be performed. Include [these criteria](https://github.com/f-dangel/backpack/blob/2c632c9fd2e54119595fc32745eaec4d3babdb4b/backpack/extensions/module_extension.py#L93-L96)

backpack/__init__.py
```diff
@@ -66,6 +74,15 @@ def __enter__(self):
...
     def __exit__(self, type, value, traceback):
+        """Leave backpack environment.
+
+        Args:
+            type: .
+            value: .
+            traceback: .
+        """
+        for backprop_extension in CTX.get_active_exts():
+            backprop_extension.clear()
```
f-dangel, line 91: Delete and test that `saved_quantities` is empty after backpropagation with `retain_grad=False`.

backpack/extensions/module_extension.py
f-dangel: `if not keep_graph: # delete backpropagated quantity on output`

backpack/extensions/module_extension.py
f-dangel: Reintroduce, then this will be resolved.

backpack/extensions/module_extension.py
schaefertim: Added the function.

backpack/__init__.py
schaefertim: Deleted. Test still missing.

schaefertim: Added the test.

backpack/extensions/module_extension.py
schaefertim: Reintroduced. But I had to comment out `inp_is_out` since it breaks the current implementation.

backpack/extensions/module_extension.py
schaefertim: Realized, but at the beginning of the `__call__` function instead of at the end.

backpack/extensions/module_extension.py
f-dangel: Do you have an idea why it breaks?

backpack/extensions/module_extension.py
```diff
@@ -12,98 +24,198 @@ class ModuleExtension:
...
+        """
+        delete_old_quantities = not self.__should_retain_backproped_quantities(module)
+        bp_quantity = self.__get_backproped_quantity(
+            extension, module.output, delete_old_quantities
+        )
+        if (
+            extension.expects_backpropagation_quantities()
+            and bp_quantity is None
+            and not is_loss(module)
+            and TORCH_VERSION_AT_LEAST_1_9_0
```
f-dangel, line 103: Can you explain to me why the `AssertionError` is only raised for specific PyTorch versions?

backpack/extensions/module_extension.py
schaefertim: Yes.
Modules like ReLU have same input and output.
Old: inp_is_out=True -> should_retain_grad=True -> gets overwritten it in the next module
New (with inp_is_out): inp_is_out=True -> should_retain_grad=True -> raises Error in nextModule: Quantitiy already exists, to the code it looks like a ResNet but ResNets are not supported yet
New (without inp_is_out): should_retain_grad=False -> quantity gets deleted

I'm not sure how to resolve the conflict between retain_grad=True and ResNets, because they both require quantities to have the same save_vector(old) or data_ptr (new).

backpack/extensions/module_extension.py
schaefertim: Because only the full_backward_hook executes the modules in exactly the same order as the forward hook.
With the old hook it can happen, that the order is changed by PyTorch. In this case, sometimes the bp_quantitiy is None.

backpack/extensions/module_extension.py
```diff
@@ -12,98 +24,198 @@ class ModuleExtension:
...
+            whether backpropagation quantities should be kept
+        """
+        is_a_leaf = module.output.grad_fn is None
+        retain_grad_is_on = getattr(module.output, "retains_grad", False)
+        # inp_is_out = id(module.input0) == id(module.output)
+        should_retain_grad = is_a_leaf or retain_grad_is_on  # or inp_is_out
+        print(f"module: {module}")
+        print(f"is_a_leaf={is_a_leaf}")
+        print(f"retain_grad_is_on={retain_grad_is_on}")
+        # print(f"inp_is_out={inp_is_out}")
```
f-dangel, line 155: Remove `print` statements before merging

backpack/extensions/module_extension.py
f-dangel: Thanks -> resolved.

backpack/extensions/module_extension.py
f-dangel: Yes.

test/extensions/test_hooks.py
schaefertim: I reintroduced these tests. Need to improve documentation (this can be discussed based on the new code).
Resolve this conversation.

backpack/extensions/module_extension.py
schaefertim: Already removed them.

backpack/extensions/module_extension.py
f-dangel: Can you post the output of this snippet for `torch>=1.9.0`?
```python
  from torch import rand
  from torch.nn import ReLU
  from backpack import extend

  a = rand(10, 10)
  layer = extend(ReLU())
  b = layer(a)

  print(id(layer.input0) == id(layer.output))
```
(I get `False` on `1.7.1`)

schaefertim: `False` on `1.9.0`

schaefertim: Seems like my description above is completely wrong and the issue is somewhere else.

f-dangel: Waiting for more information. Maybe requires in-person discussion

backpack/utils/__init__.py
```diff
@@ -1,8 +1,26 @@
...
+    Args:
+        error: previous exception type
+
+    Returns:
+        new exception type
+    """
+    if TORCH_VERSION_AT_LEAST_1_7_0:
+        return error
+    else:
+        return RuntimeError
```
schaefertim, line 26: Note: I introduced this function to have a single interface for this kind of behaviour. It seemed like someone previously encountered the same problem. So I replaced it with this function.

backpack/extensions/module_extension.py
schaefertim: The above statement holds, but for `Flatten` instead of `ReLU`.
```Python
from torch import rand
from torch.nn import Flatten
from backpack import extend

a = rand(10, 10)
layer = extend(Flatten())
b = layer(a)

print(id(layer.input0) == id(layer.output))
```
This is `True` only if `Flatten` does nothing. Change `a=rand(10, 10, 10)` and the ids are different.

backpack/extensions/saved_quantities.py
```diff
@@ -0,0 +1,46 @@
...
+                "quantities like in ResNets are not supported yet."
+            )
+        else:
+            self._saved_quantities[key] = quantity
+
+    def retrieve_quantity(self, key: int, delete_old: bool) -> Union[Tensor, None]:
+        """Returns the saved quantity.
+
+        Args:
+            key: id of grad_output[0]
```
f-dangel, line 37: is this correct?

backpack/utils/__init__.py
f-dangel: Okay.

backpack/extensions/module_extension.py
f-dangel: I think I understand the problem:

If `Flatten` does not perform an operation, it should simply retain the backpropagated quantity and skip its own `backpropagate` method. 

Your current code does not retain the backpropagated quantity and calls `Flatten`'s `backpropagate` function.

I tried out the first approach in [this commit](https://github.com/f-dangel/backpack/commit/668e3aa343075da17b163b7e09b521312deec768) and was able to get rid of a lot of additional boilerplate in `Flatten`'s second-order extensions. Can you take a look at the diff and try to spot any flaws? If it looks good, you can merge it into your branch.

backpack/extensions/saved_quantities.py
schaefertim: Yes, the key should be the id of grad_output[0] when retrieving the quantitiy.

backpack/extensions/module_extension.py
f-dangel: Thought a bit more about it. Approach 2 also implies that `backpropagate` of activation functions with `inplace=True`, like `Relu(inplace=True)` will not be called because `id(input0) == id(output)`. But in this case an exception should be raised that such inplace operations are not supported by BackPACK. Hence, approach 2 could lead to situations where BackPACK computes wrong results but the code passes.

Having thought about those edge cases, I think we can stay with your current (1.) solution and leave `inp_is_out` commented out.

Marking as resolved. Reopen in case of follow-up questions.

test/extensions/graph_clear_test.py
```diff
@@ -0,0 +1,61 @@
...
+
+    Raises:
+        NotImplementedError: if problem string is unknown
+    """
+    batch_size, in_dim, out_dim = 2, 3, 4
+    inputs = rand(batch_size, in_dim)
+    if request.param == PROBLEM_STRING[0]:
+        model = Sequential(Linear(in_dim, out_dim), ReLU(), Linear(out_dim, out_dim))
+    elif request.param == PROBLEM_STRING[1]:
+        model = Sequential(Linear(in_dim, out_dim), Flatten(), Linear(out_dim, out_dim))
```
schaefertim, line 53: Tests whether the graph is clear after the backward pass.
Test setting: flatten_no_op
Passes for torch 1.9.0. Fails for low versions.

schaefertim: Fixed it in module_extension.py by skipping Flatten if no_op.

I would be interested in the test coverage. In the Flatten module extensions, the lines overwriting backpropagate, can maybe be deleted now, depending on whether they are still called.

backpack/extensions/saved_quantities.py
schaefertim: Totally wrong by me. Of course it is not the id but the data_ptr.
Corrected it.

test/extensions/graph_clear_test.py
f-dangel: Improved readability a bit an included an additional check that makes sure the `Flatten` was indeed a no op.

f-dangel: Could remove the logic inside the second-order `Flatten` module extensions.

f-dangel:
Coverage report: Gradient extension for `BatchNormNd` is [untested](https://coveralls.io/builds/41785113/source?filename=backpack%2Fextensions%2Ffirstorder%2Fgradient%2Fbatchnorm_nd.py)

f-dangel:
Here are my notes what is changing with this PR:

---

-   Use `register_full_backward_hook` for `torch>=1.9.0`, else `register_backward_hook`: `full_backward_hook` is already introduced in `torch==1.8.0`, but it won't be fired on modules whose input has `requires_grad=False` (e.g. first layer of a net) before `torch==1.9.0`.
-   Modify which modules are extended:
    -   Container modules built-in to PyTorch: **Old:** Have hooks that store I/O in the forward, and free I/O in the backward pass. **New:** Don't have hooks.
    -   As a consequence of the above: **Old:** Post extension hooks can be executed on built-in PyTorch containers and their submodules. **New:** Post extension hooks only execute on submodules of built-in containers.
-   Modify backpropagation for second-order quantities:
    -   **Old:** Call `backpropagate` from a module extension even if `module.input0` is end of backpropagation (does not require gradient). **New:** Only backpropagate if `module.input0` requires gradient.
    -   Storage of additional backpropagated information: **Old:** Fetch information from `module.output`, backpropagate through module and append result to `module.input0`. This is not possible anymore with the `full_backward_hook` because of <https://discuss.pytorch.org/t/register-full-backward-hook-alters-behaviour-of-forward-hook/123466> (information attached to python objects is lost). **New:** Keep a separate dictionary that uses the `data_ptr()` of `module.output` and `module.input0` as identifier for `output` and `input`. This property is preserved, i.e. `layer2.input0.data_ptr() == layer1.output.data_ptr()`.
    -   `Flatten` layers that act as no op: **Old:** Second-order module extensions for `Flatten` check if it was a no op and the hook is therefore called at the wrong time. **New:** With `register_full_backward_hook` the execution order gets fixed. With `register_backward_hook` check in module extension whether the execution order is wrong and skip backpropagation only if the `Flatten` module was a no op. No extra logic in the `Flatten` second-order module extensions.

---

@schaefertim  anything important I am missing here?

schaefertim:
> @schaefertim anything important I am missing here?

No, nothing missing. These are the major changes.

backpack/extensions/saved_quantities.py
```diff
@@ -0,0 +1,50 @@
...
+        """Initialization."""
+        self._saved_quantities: Dict[int, Tensor] = {}
+
+    def save_quantity(self, key: int, quantity: Tensor) -> None:
+        """Saves the quantity under the specified key.
+
+        Args:
+            key: data_ptr() of reference tensor.
+                For torch>=1.9.0 the reference tensor is grad_input[0].
+                For older versions the reference tensor is module.input0.
```
f-dangel, line 20: I think this is still incorrect (see [this line](https://github.com/tjs111/backpack/blob/fa09668cdc043982b89ac395ca77ef001a281a57/backpack/extensions/module_extension.py#L134). We always use `module.input0.data_ptr()` as identifier.

f-dangel: Pushing a correction. Let me know if I missed something.

PR #199, opened 2021-07-06
[TEST] Reduce run time

f-dangel:
The final layers of automated test settings can have large final linear layers
that make the computation of second-order quantities expensive. Disabling 
their `requires_grad` speeds up the tests.

This PR:

- makes the automated settings' last linear layers non-differentiable
- updates the new test suite, as it assumes all model parameters are differentiable
- updates `ggn_vector_product` to work with models that contain non-differentiable
   parameters and fully-documents `ggnvp.py`

@schaefertim Any concerns? I will add more documentation in the test suite before merging.


schaefertim:
> @schaefertim Any concerns? I will add more documentation in the test suite before merging.

I like the idea a lot and it seems to save a significant part of the runtime.
I have one suggestion, that is based on my concern that my new hook might not comply with this (yet). At the moment the test just collects the data from the `model.trainable_parameters()`. I would prefer if the test checked explicitly if only the trainable parameters have gradients. Something along these lines
```
def _collect_data(model: Model, savefield: str) -> List[Tensor]:
    data: List[Tensor] = []
    for p in model.parameters():
        if p.requires_grad:
            data.append(getattr(p, savefield))
        else:
            if hasattr(p, savefield):
                raise AssertionError("There must be no gradients for non-trainable parameters")
    return data
```

f-dangel:
Incorporated your suggestion and added documentation.

PR #211, opened 2021-08-06
[ADD] CrossEntropyLoss: support additional axes

schaefertim:
CrossEntropyLoss supports additional dimensions, which can be seen from the [PyTorch Docs](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html).
This adds support for those additional dimensions.

Also adds a test for `make_hessian_mat_prod`.


test/core/derivatives/loss_settings.py
```diff
@@ -37,15 +37,25 @@
...
         "module_fn": lambda: torch.nn.CrossEntropyLoss(reduction="mean"),
         "input_fn": lambda: torch.rand(size=(2, 4)),
-        "target_fn": lambda: classification_targets(size=(2,), num_classes=2),
+        "target_fn": lambda: classification_targets(size=(2,), num_classes=4),
     },
     {
         "module_fn": lambda: torch.nn.CrossEntropyLoss(reduction="sum"),
         "input_fn": lambda: torch.rand(size=(8, 4)),
-        "target_fn": lambda: classification_targets(size=(8,), num_classes=2),
+        "target_fn": lambda: classification_targets(size=(8,), num_classes=4),
```
f-dangel, line 58: Why does `num_classes` change?

test/core/derivatives/loss_settings.py
```diff
@@ -28,7 +28,7 @@
 example = {
     "module_fn": lambda: torch.nn.CrossEntropyLoss(reduction="mean"),
     "input_fn": lambda: torch.rand(size=(2, 4)),
-    "target_fn": lambda: classification_targets(size=(2,), num_classes=2),
+    "target_fn": lambda: classification_targets(size=(2,), num_classes=4),
```
f-dangel, line 31: Why does `num_classes` change?

backpack/core/derivatives/crossentropyloss.py
```diff
@@ -64,49 +69,57 @@ def _sqrt_hessian_sampled(
         sqrt_mc_h = (probs_unsqueezed - classes) / sqrt(M)
 
         if module.reduction == "mean":
-            N = module.input0.shape[0]
-            sqrt_mc_h /= sqrt(N)
+            sqrt_mc_h /= sqrt(self._get_number_of_samples(module, probs, subsampling))
```
f-dangel, line 72: same remark about the normalization as above.

backpack/core/derivatives/crossentropyloss.py
```diff
@@ -64,49 +69,57 @@ def _sqrt_hessian_sampled(
...
-            Hmat = einsum("bi,cbi->cbi", (probs, mat)) - einsum(
-                "bi,bj,cbj->cbi", (probs, probs, mat)
+            Hmat = einsum("bi,cbi->cbi", probs, mat) - einsum(
+                "bi,bj,cbj->cbi", probs, probs, mat
             )
 
             if module.reduction == "mean":
-                N = module.input0.shape[0]
-                Hmat /= N
+                Hmat /= self._get_number_of_samples(module, probs, subsampling=None)
```
f-dangel, line 106: same remark about normalization as above.

backpack/core/derivatives/crossentropyloss.py
```diff
@@ -145,3 +158,75 @@ def _check_2nd_order_parameters(self, module):
...
+        else:
+            str_d_dims = ""
+            d_info = {}
+        return probs, input_dim, str_d_dims, d_info
+
+    @staticmethod
+    def _rearrange_output(tensor: Tensor, input_dim, str_d_dims, d_info) -> Tensor:
+        """Rearrange the output. Used together with rearrange_input.
+
+        If input_dim>=3: rearrange c1 (n d1 d2) c2 -> c1 n c2 d1 d2
```
f-dangel, line 194: Rename `c1 -> v`, then `c2 -> c1`?

backpack/core/derivatives/crossentropyloss.py
```diff
@@ -145,3 +158,75 @@ def _check_2nd_order_parameters(self, module):
...
+            input_dim: the number of input dimensions
+            str_d_dims: a string representation of the additional dimensions
+            d_info: a dictionary encoding the size of the additional dimensions
+
+        Returns:
+            the rearranged tensor
+        """
+        if input_dim >= 3:
+            tensor = rearrange(
+                tensor, f"c1 (n {str_d_dims}) c2 -> c1 n c2 {str_d_dims}", **d_info
```
f-dangel, line 207: same renaming as above

backpack/core/derivatives/crossentropyloss.py
```diff
@@ -24,20 +25,22 @@ def _sqrt_hessian(
...
         tau = torchsqrt(probs)
         V_dim, C_dim = 0, 2
         Id = diag_embed(ones_like(probs), dim1=V_dim, dim2=C_dim)
         Id_tautau = Id - einsum("nv,nc->vnc", tau, tau)
         sqrt_H = einsum("nc,vnc->vnc", tau, Id_tautau)
 
         if module.reduction == "mean":
-            N = module.input0.shape[0]
-            sqrt_H /= sqrt(N)
+            sqrt_H /= sqrt(self._get_number_of_samples(module, probs, subsampling))
```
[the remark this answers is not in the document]
f-dangel: My bad, the function actually computes this number. Refactoring it to make it clearer.

test/core/derivatives/loss_settings.py
schaefertim: Because, the input is (N, C) with batch size N and number of classes C. This correction puts the targets in line with the input.

test/core/derivatives/loss_settings.py
schaefertim: Because, the input is (N, C) with batch size N and number of classes C. This correction puts the targets in line with the input.

f-dangel: Makes sense.

backpack/core/derivatives/crossentropyloss.py
```diff
@@ -64,49 +69,56 @@ def _sqrt_hessian_sampled(
...
         self._check_2nd_order_parameters(module)
 
         probs = self._get_probs(module)
+        probs, *rearrange_info = self._merge_batch_and_additional(probs)
 
         def hessian_mat_prod(mat):
-            Hmat = einsum("bi,cbi->cbi", (probs, mat)) - einsum(
-                "bi,bj,cbj->cbi", (probs, probs, mat)
+            Hmat = einsum("bi,cbi->cbi", probs, mat) - einsum(
+                "bi,bj,cbj->cbi", probs, probs, mat
```
f-dangel, line 101: `mat.dim()` is equal to `input.dim() + 1`. This `einsum` will fail unless `input.dim() == 2`. 

The problem for this PR is that this function is currently only tested by the old test suite (2d inputs to cross-entropy loss). The new test suite does not implement tests for this function. It would be good to add them.

schaefertim: outdated

backpack/core/derivatives/crossentropyloss.py
```diff
@@ -1,9 +1,13 @@
 """Partial derivatives for cross-entropy loss."""
+from itertools import product as itertools_product
 from math import sqrt
-from typing import List, Tuple
+from typing import Callable, Dict, Iterable, List, Tuple, Union
 
-from torch import Tensor, diag, diag_embed, einsum, multinomial, ones_like, softmax
+from einops import rearrange
+from numpy import prod
```
schaefertim, line 7:
```suggestion
from math import prod
```
This would be nicer but doesn't work. (Import error)

backpack/core/derivatives/crossentropyloss.py
```diff
@@ -64,49 +73,73 @@ def _sqrt_hessian_sampled(
...
+                *probs.shape[1:], *probs.shape[1:]
+            )
+            sum_H = (
+                diagonal
+                - einsum(
+                    "ncx,ndy,xy->cxdy",
+                    probs.flatten(2),
+                    probs.flatten(2),
+                    eye(
+                        prod(probs.shape[2:]),
```
f-dangel, line 102: I think you can do `probs.shape[2:].numel()` and get rid of the `prod` import

backpack/core/derivatives/crossentropyloss.py
```diff
@@ -145,3 +178,120 @@ def _check_2nd_order_parameters(self, module):
...
+                selection_sqrt_H = (slice(None),) * 3 + index_additional_axes
+                selection_target = (
+                    (slice(None),)
+                    + index_additional_axes
+                    + (slice(None),) * 2
+                    + index_additional_axes
+                )
+                sqrt_H_full[selection_target] = sqrt_h[selection_sqrt_H]
+            return sqrt_H_full.flatten(0, n_additional_axes)
+        return sqrt_h
```
f-dangel, line 285: I think that, based on Tuesday's write-up, there should be a way to implement this cleaner, using [`torch.diag_embed`](https://pytorch.org/docs/stable/generated/torch.diag_embed.html) and no `for` loops.

backpack/core/derivatives/crossentropyloss.py
schaefertim: Yes, that works. Thank you for this solution.

backpack/core/derivatives/crossentropyloss.py
schaefertim: I also feel like there should be a way. But after thinking for a while, I don't think there is a way.
`torch.diag_embed` does a 2D embedding. Since this is a multidimensional embedding, it would have to be a concatenation of 2D-diagonal embeddings. Do you think this is faster than the for-loop? (My guess is, that it does just the same)

f-dangel: The documentation says that (d>1)-dimensional inputs can also be embedded. My guess is that if we can use `diag_embed` and some `reshape`s, this should be faster than a for loop.

schaefertim: Indeed, this works.
Using flatten - diag_embed - reshape.

f-dangel:
Everything looking good now in `core`.

Please add a test case for `DiagGGN{Exact, MC}` and `SqrtGGN{Exact, MC}` extensions.

schaefertim:
> Everything looking good now in `core`.
> 
> Please add a test case for `DiagGGN{Exact, MC}` and `SqrtGGN{Exact, MC}` extensions.

Done.

f-dangel:
Can you resolve the merge conflict?

schaefertim:
> Can you resolve the merge conflict?

Done.

f-dangel:
All good here. Waiting for `coveralls` to be back online ([status](https://status.coveralls.io/))

Issue #213, opened 2021-08-19
Second order gradient for Batch-wise HVP


haonan3:
Hi,

Thanks for this great library. I am wondering how can we compute the gradient of sum of gradient?

I am trying to implement the Hessian-Vector-Product (HVP) with the following code:

```
def batch_hvp(self, model, loss, params_list, batch_grad_list):
    if len(params_list) != len(batch_grad_list):
        raise (ValueError("w and v must have the same length."))

    one_sample_grad_list = grad(loss, params_list, retain_graph=True, create_graph=True)

    elemwise_products = 0
    for grad_elem, v_elem in zip(one_sample_grad_list, batch_grad_list):
        sum_over_dims = []
        for i in range(len(v_elem.shape)):
            sum_over_dims.append(i)
        sum_over_dims = tuple(sum_over_dims[1:])
        elemwise_products += torch.sum(grad_elem.unsqueeze(0) * v_elem.detach(), sum_over_dims)

    with backpack(BatchGrad()):
        elemwise_products.backward()   # problem: has no attribute 'input0'
        return_grads = [p.grad_batch for p in model.parameters() if p.requires_grad]

    return return_grads
```
I encounter the  " has no attribute 'input0'  " problem when I call the backward(), is it possible to get batch gradients for return_grads?

For now, I am only using the for loop to compute the gradient.

```
def batch_hvp(self, model, loss, params_list, batch_grad_list):
    if len(params_list) != len(batch_grad_list):
        raise (ValueError("w and v must have the same length."))

    one_sample_grad_list = grad(loss, params_list, retain_graph=True, create_graph=True)

    elemwise_products = 0
    for grad_elem, v_elem in zip(one_sample_grad_list, batch_grad_list):
        sum_over_dims = []
        for i in range(len(v_elem.shape)):
            sum_over_dims.append(i)
        sum_over_dims = tuple(sum_over_dims[1:])
        elemwise_products += torch.sum(grad_elem.unsqueeze(0) * v_elem.detach(), sum_over_dims)
    
    # The for-loop version 
    grad_cache = []
    for i in range(elemwise_products.shape[0]):
        elemwise_products[i].backward(retain_graph=True)
        grad_cache.append([p.grad.clone() for p in model.parameters() if p.requires_grad])
    grad_cache = list(zip(*grad_cache))
    return_grads = []
    for l_id in range(len(grad_cache)):
[...]

PR #214, opened 2021-08-30
retain_graph=True (search variable in stack)

schaefertim:
Handle the case where `retain_graph=True`.
This fixes https://github.com/f-dangel/backpack/issues/213
and fixes first half of https://github.com/fKunstner/backpack-discuss/issues/115

This solution was suggested to me in https://discuss.pytorch.org/t/infer-value-of-retain-graph-and-create-graph/130622

Runtime overhead:
A noticed a significant runtime overhead: the test runtime increased from ~5min to ~18min.
In [this stackoverflow post](https://stackoverflow.com/questions/17407119/python-inspect-stack-is-slow), the `inspect.stack()` call had a runtime of 5e-3s. 
Using `inspect.stack(0)` instead, reduced the overhead by a factor of 5. For our case, tests would then run in ~7.6 min.
The other suggestion is to use a recursive scheme, that only inspects the previous frame via `frame.f_back`. I'm not sure how many recursive calls would be needed to reach the frame that is `autograd`.

CUDA:
This function does not work on CUDA!

Convert to draft. Eventually close, unless there is a new idea coming up.

### DIFF
--- a/backpack/__init__.py
+++ b/backpack/__init__.py
@@ -205,7 +205,10 @@ def hook_run_extensions(
         print("[DEBUG] Running extension hook on", module)
     CTX.get_extension_hook()(module)
 
-    if not (
+    # could be made more precise to check for backward/grad
+    fr = [f[0] for f in inspect.stack() if "/torch/autograd/" in f.filename][-1]
+    retain_graph = inspect.getargvalues(fr).locals["retain_graph"]
+    if not retain_graph and not (
         CTX.is_extension_active(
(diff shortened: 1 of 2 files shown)


f-dangel:
I agree with [the above post](https://discuss.pytorch.org/t/infer-value-of-retain-graph-and-create-graph/130622/2?u): This solution can break easily in the future.

Favoring #217, which introduces a `retain_graph` argument to BackPACK's context manager.

PR #215, opened 2021-08-31
[ADD] LSTM: batch_first, extensions, tests

schaefertim:
LSTM module:
- allow `batch_first=True`
- extensions: first order extensions, DiagGGN
- tests for derivatives and extensions


backpack/extensions/firstorder/variance/rnn.py
```diff
@@ -16,5 +18,21 @@ def __init__(self):
...
+        """Initialization."""
+        super().__init__(
+            params=["bias_ih_l0", "bias_hh_l0", "weight_ih_l0", "weight_hh_l0"],
+            grad_extension=GradLSTM(),
+            sgs_extension=SGSLSTM(),
+        )
+
+    @staticmethod
+    def _get_axis_batch(module: LSTM) -> int:
+        return 0 if module.batch_first else 1
```
f-dangel, line 38: Use [`get_batch_axis`](https://github.com/tjs111/backpack/blob/458c32038255fdfc9d830570e286b2e9956218c3/backpack/utils/subsampling.py#L28-L65). Same for other occurrences.

f-dangel: Done.

backpack/core/derivatives/lstm.py
```diff
@@ -349,14 +326,13 @@ def _weight_ih_l0_jac_t_mat_prod(
             module, mat, subsampling=subsampling
         )
 
+        N_axis = get_batch_axis(module, "input0")
```
schaefertim, line 329:
```suggestion
        N_axis, _ = self._get_batch_and_time_axes(module)
```
maybe it's better to use the method from the same class

backpack/core/derivatives/lstm.py
```diff
@@ -370,29 +346,45 @@ def _weight_hh_l0_jac_t_mat_prod(
...
+        Args:
+            module: LSTM module.
+
+        Returns:
+            Batch axis and time axis.
+        """
+        N_axis = get_batch_axis(module, "input0")
+        T_axis = 1 if N_axis == 0 else 0
+
+        return N_axis, T_axis
```
schaefertim, line 390:
```suggestion
        return N_axis, T_axis = (0, 1) if module.batch_first else (1, 0)
```

backpack/core/derivatives/lstm.py
f-dangel: Agree. Also:
- [ ] call `get_batch_and_time_axes` from `get_batch_axis`

schaefertim: I agree that it would be nice to change `get_batch_axis`.
However, this would lead to a cyclic import (`subsampling` imports `lstm.get_batch_and_time_axes` and `lstm` imports `subsampling.subsample`).
Solution would be to move `get_batch_axis`()` to a separate file (e.g. `batch_axis.py`). This would be a better sorting of functions but involves many files.

backpack/core/derivatives/lstm.py
schaefertim: Applied this.

backpack/core/derivatives/lstm.py
f-dangel: Thanks for summing up the options. There would be a significant diff if we move `get_batch_axis`.

PR #216, opened 2021-08-31
[ADD] Embedding: derivatives, extensions, tests

schaefertim:
Embedding:
- Derivatives: only derivative wrt weight, since derivative wrt input not well-defined
- Extensions: first order, DiagGGN
- tests: derivatives, extensions

- [x] raise similar Error: `RuntimeError: only Tensors of floating point and complex dtype can require gradients`


f-dangel:
**Question:** Currently, BackPACK assumes `batch_axis=0` for any `Embedding` layer. For cases with batch-second, this will lead to wrong results. How can this be addressed?

test/core/derivatives/problem.py
```diff
@@ -152,7 +152,8 @@ def forward_pass(
             input = subsample(input, dim=batch_axis_in, subsampling=subsampling)
 
         if input_requires_grad:
-            input.requires_grad = True
+            if input.dtype is float32:
+                input.requires_grad = True
```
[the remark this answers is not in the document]
schaefertim: `RuntimeError: only Tensors of floating point and complex dtype can require gradients` is the error message if you try to assign `input.requires_grad=True` on long.
So yes, your suggestion would work fine. 
Do you think this is better? If yes, I'll change it.

schaefertim:
> **Question:** Currently, BackPACK assumes `batch_axis=0` for any `Embedding` layer. For cases with batch-second, this will lead to wrong results. How can this be addressed?

I don't think there is a way to determine this from within BackPACK. The only ways I can think of include graph analysis which is way too much effort.

f-dangel:
> > **Question:** Currently, BackPACK assumes `batch_axis=0` for any `Embedding` layer. For cases with batch-second, this will lead to wrong results. How can this be addressed?
> 
> I don't think there is a way to determine this from within BackPACK. The only ways I can think of include graph analysis which is way too much effort.

That is a problem for all extensions that compute per-sample quantities. I am currently not sure how this should best be resolved.

test/core/derivatives/problem.py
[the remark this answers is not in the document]
f-dangel: I would prefer it because it's less strict. We don't use complex tensors in BackPACK at the moment, but if we did, your code will not set `requires_grad`. At the same time, making the check more complicated by also checking whether `dtype` is complex makes the code unnecessarily complicated, because there are no complex tensors in BackPACK at the moment.

f-dangel:
> > > **Question:** Currently, BackPACK assumes `batch_axis=0` for any `Embedding` layer. For cases with batch-second, this will lead to wrong results. How can this be addressed?
> > 
> > 
> > I don't think there is a way to determine this from within BackPACK. The only ways I can think of include graph analysis which is way too much effort.
> 
> That is a problem for all extensions that compute per-sample quantities or support mini-batch sub-sampling. I am currently not sure how this should best be resolved.

As discussed offline, BackPACK will require access to the batch axis of an `Embedding` module in most cases:
- `Embedding` modules must have a `.batch_axis` argument that must be either `0` or `1` and indicate which axis of the input is the batch axis
- If not provided, BackPACK will raise an Exception that explains the above point. The exception will only be raised in the extensions/features that require access to the batch axis

test/core/derivatives/problem.py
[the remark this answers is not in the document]
schaefertim: Changed condition to `if input.dtype is not long`

f-dangel:
As discussed offline, for now `Embedding` is assuming `batch_axis=0`.

backpack/core/derivatives/embedding.py
```diff
@@ -0,0 +1,76 @@
...
+        for s in range(module.num_embeddings):
+            delta[s] = input0 == s
+        if TORCH_VERSION_AT_LEAST_1_9_0:
+            equation = f"sn...,vn...h->v{'' if sum_batch else 'n'}sh"
+        elif delta.dim() >= 3:
+            equation = f"snx,vnxh->v{'' if sum_batch else 'n'}sh"
+            delta = flatten(delta, start_dim=2, end_dim=-1)
+            mat = flatten(mat, start_dim=2, end_dim=-2)
+        else:
+            equation = f"sn,vnh->v{'' if sum_batch else 'n'}sh"
```
f-dangel, line 62: Should this be in an `else` branch and further indentet? Because now if the equation with ellipsis can be used, but `delta.dim() >= 3`, then `delta` and `mat` will be flattened, altough this is not necessary.

backpack/core/derivatives/embedding.py
```diff
@@ -0,0 +1,76 @@
...
+        input0 = subsample(module.input0, subsampling=subsampling)
+        delta = zeros(module.num_embeddings, *input0.shape, device=mat.device)
+        for s in range(module.num_embeddings):
+            delta[s] = input0 == s
+        if TORCH_VERSION_AT_LEAST_1_9_0:
+            equation = f"sn...,vn...h->v{'' if sum_batch else 'n'}sh"
+        elif delta.dim() >= 3:
+            equation = f"snx,vnxh->v{'' if sum_batch else 'n'}sh"
+            delta = flatten(delta, start_dim=2, end_dim=-1)
+            mat = flatten(mat, start_dim=2, end_dim=-2)
```
f-dangel, line 60: use `delta.flatten(...)`, `mat.flatten(...)`

backpack/core/derivatives/embedding.py
```diff
@@ -0,0 +1,76 @@
...
+        return einsum(equation, delta, mat)
+
+    def _check_parameters(self, module: Embedding) -> None:
+        if module.padding_idx is not None:
+            raise NotImplementedError("Only padding_idx=None supported.")
+        elif module.max_norm is not None:
+            raise NotImplementedError("Only max_norm=None supported.")
+        elif module.scale_grad_by_freq:
+            raise NotImplementedError("Only scale_grad_by_freq=False supported.")
+        elif module.sparse:  # TODO sparse might be supported -> test
```
f-dangel, line 72: Update the TODO

backpack/core/derivatives/embedding.py
```diff
@@ -0,0 +1,76 @@
...
+            raise NotImplementedError("Only sparse=False supported.")
+
+    def hessian_is_zero(self, module: Embedding) -> bool:  # noqa: D102
+        return False  # TODO discuss
```
f-dangel, line 76: Still requires discussion? I think the same argument as for `_jac_t_mat_prod` applies

backpack/core/derivatives/embedding.py
```diff
@@ -0,0 +1,74 @@
...
+    ) -> Tensor:
+        self._check_parameters(module)
+
+        input0 = subsample(module.input0, subsampling=subsampling)
+        delta = zeros(module.num_embeddings, *input0.shape, device=mat.device)
+        for s in range(module.num_embeddings):
+            delta[s] = input0 == s
+        if TORCH_VERSION_AT_LEAST_1_9_0:
+            equation = f"sn...,vn...h->v{'' if sum_batch else 'n'}sh"
+        elif delta.dim() >= 3:
```
f-dangel, line 55: `>3` sufficient?

test/extensions/secondorder/diag_ggn/diag_ggn_settings.py
```diff
@@ -147,6 +148,32 @@
...
+    },
+    {
+        "input_fn": lambda: randint(0, 3, (3, 2, 2)),
+        "module_fn": lambda: Sequential(
+            Embedding(3, 2),
+            Flatten(),
+        ),
+        "loss_function_fn": lambda: CrossEntropyLoss(reduction="mean"),
+        "target_fn": lambda: classification_targets((3,), 2 * 2),
+        "seed": 1,
```
f-dangel, line 172: can this line be removed?

backpack/core/derivatives/embedding.py
schaefertim: I think it is fine as is.

If ellipsis can be used (torch >= 1.9.0) all the other stuff is jumped.
For clarification, maybe the two branches should be swapped and the if-condition changed to `if delta.dim() == 2`. Is this better?

backpack/core/derivatives/embedding.py
schaefertim: Done.

backpack/core/derivatives/embedding.py
schaefertim: sparse does not work -> remove TODO

backpack/core/derivatives/embedding.py
schaefertim: I agree.
I'll raise the same error.

test/extensions/secondorder/diag_ggn/diag_ggn_settings.py
schaefertim: No, is necessary.

backpack/core/derivatives/embedding.py
schaefertim: Solved it differently.
I removed the last else branch.

backpack/core/derivatives/embedding.py
schaefertim: Yes, but solved it differently. See https://github.com/f-dangel/backpack/pull/216#discussion_r711948833

backpack/core/derivatives/embedding.py
f-dangel: You're right, my interpretation of the code was wrong.

If like the proposed change of the `if` condition.

test/extensions/secondorder/diag_ggn/diag_ggn_settings.py
f-dangel: Reason? Are the tolerances of `allclose` to strict with `seed=0`?

schaefertim: from seeds 0-9 three seeds fail: 0, 4, 9. For all of them, the `diag_ggn_mc[_batch]_light` test on cuda fails. For seed 0 additionally the `diag_ggn_mc_batch_light` test on cpu fails.
I'm not sure what the reason is. The values are off by approximately 10%-30%.

f-dangel: Could you quickly verify if these seeds still fail if you
- append a zero to `mc_samples` [here](https://github.com/tjs111/backpack/blob/5abe431b8f093e7b87643a50f9989ae4a9918196/test/extensions/secondorder/diag_ggn/test_diag_ggn.py#L50)
- append a zero to both `mc_samples` and `chunks` [here](https://github.com/tjs111/backpack/blob/5abe431b8f093e7b87643a50f9989ae4a9918196/test/extensions/secondorder/diag_ggn/test_diag_ggn.py#L72-L73)
(maybe you have to do the same for `BatchDiagGGNMC`

schaefertim: Adding a zero to `mc_samples` ([here](https://github.com/schaefertim/backpack/blob/5abe431b8f093e7b87643a50f9989ae4a9918196/test/extensions/secondorder/diag_ggn/test_diag_ggn.py#L50) and [here](https://github.com/schaefertim/backpack/blob/5abe431b8f093e7b87643a50f9989ae4a9918196/test/extensions/secondorder/diag_ggn/test_batch_diag_ggn.py#L50)) would resolve this.

f-dangel: Fine, then `0, 4, 9` are just 'unlucky' seeds. Thanks for double checking. Resolving.

backpack/core/derivatives/embedding.py
```diff
@@ -0,0 +1,72 @@
...
+
+        input0 = subsample(module.input0, subsampling=subsampling)
+        delta = zeros(module.num_embeddings, *input0.shape, device=mat.device)
+        for s in range(module.num_embeddings):
+            delta[s] = input0 == s
+        if TORCH_VERSION_AT_LEAST_1_9_0:
+            equation = f"sn...,vn...h->v{'' if sum_batch else 'n'}sh"
+        else:
+            equation = f"snx,vnxh->v{'' if sum_batch else 'n'}sh"
+            delta = delta.flatten(start_dim=2, end_dim=-1)
```
f-dangel, line 57: tests fail here

schaefertim: my bad. 
Fixed it by reintroducing the if-else branch. Now the condition is `if delta.dim() == 2`.

f-dangel:
All good here. Waiting for `coveralls` to be back online ([status](https://status.coveralls.io/))

f-dangel:
@schaefertim Could you also add the `SqrtGGN` extension for `Embedding` + tests?

Issue #239, opened 2022-01-26
Parameter `grad`s don't get initialized with `BatchL2Grad` and BatchNorm


thomasahle:
BatchL2Grad, perhaps naturally, raises an error when it sees a BatchNorm, since batch normalization mixes gradients in a way that makes the individual contribution hard to discern.
The error says I can ignore it, if I know what I'm doing. I can't say I completely do, but if I ignore it, I do indeed get both `grad`s and `batch_l2`s on the top levels of my mode, which aren't using batch-norm.
I'm happy with that.

My problem is that the lower level parameters - which do use batch norm - don't just have a None `batch_l2`, but also a None `grad`.
So my model doesn't train at all.
This seems wrong, since `grad` is indeed computable, as witnessed by PyTorch being able to do so fine without backpack.

Is there a way I can get `batch_l2`s on as many of my parameters as possible, but `grad`s on everything?

I an do this now by first calling `backward()` without backpack, and then calling it again inside `with backpack(BatchL2Grad()):`, but that seems wasteful.

f-dangel:
Hi @thomasahle,

I'm not sure I fully understand your question. Individual gradients don't exist in a network with BN layer in train mode, because the individual losses depend on all samples in the mini-batch. The purpose of the warning you're seeing is exactly to point out this caveat.

Best,
Felix

thomasahle:
Hi Felix, I'm indeed interested in individual gradient l2 norms.
If I can't get them on all my parameters, that's fine.

But it would be nice if the normal _non batched_ `grad` would still be computed for my parameters, without having to run backprop again without backpack.

Right now the non batched `grad` is only computed whenever `batch_l2` is computed.

f-dangel:
Hi Thomas,

> Right now the non batched `grad` is only computed whenever `batch_l2` is computed.

That seems odd to me because BackPACK does not intervene into PyTorch's gradient computation. Are you sure that these parameters have `requires_grad = True`?

> I'm indeed interested in individual gradient l2 norms.
If I can't get them on all my parameters, that's fine.

For the loss of a neural network with batch normalization, individual gradients, and hence their l2 norm, don't exist. BackPACK only detects this when it encounters a batch norm module. So the result in `batch_l2` in the layers before is not a per-sample gradient l2 norm. Maybe [this post I wrote](https://f-dangel.github.io/about/posts/batch_normalization.html#Batchnormalization) is helpful to understand this in more detail.

thomasahle:
[...]

PR #240, opened 2022-02-10
Better error messages for BatchNorm

fKunstner:
Makes the BatchNorm error message more explicit to avoid confusions (https://github.com/f-dangel/backpack/issues/239) and adds an option to ignore the exception.  

Summary of changes:

- Make BatchNorm error message [more explicit](https://github.com/f-dangel/backpack/commit/075834e64dff9f9297c75b4d5b51b9fcc3b5c405#diff-3b87a038266d15ca9769605a81763176d9670389ad809bea3771d43c8cc72c16R23-R28)
- Tie the BatchNorm raising of exception or warning [to fail_mode](https://github.com/f-dangel/backpack/commit/07dbbbfbab1d34b21a3d65aeb20a0da1c0bfefb9#diff-3b87a038266d15ca9769605a81763176d9670389ad809bea3771d43c8cc72c16R36-R41)
- Split error handling on missing extensions for first and second order
The current [error handling for missing modules](https://github.com/f-dangel/backpack/commit/73fd638c057c250a35e1cc1d6cd1a79325196a85#diff-cfc489564c02871bb06382d841a8a380544dce613ac05704deeb3cb31fc3a836L97-L111) mixed second-order extensions (which should fail if there is no extension defined) and first-order extensions (which should fail if there is no extension defined and the module has parameters). Moved to [first order](https://github.com/f-dangel/backpack/commit/73fd638c057c250a35e1cc1d6cd1a79325196a85#diff-9ad3a8ba51361c2bca2d17d5b7af5dd20320f1a23929f1563984775f0a7eaeebR33-R48) and [second order](https://github.com/f-dangel/backpack/commit/73fd638c057c250a35e1cc1d6cd1a79325196a85#diff-c05a11929412a47e0a2fe36222b048bae7a227e6e5b5625dd5f0ba102ecf4d75R13-R26) base classes.
- Change default fail_mode to error and make fail_mode user accessible
  First order extensions did not expose `fail_mode` and [had `warn` as a default](https://github.com/f-dangel/backpack/commit/c69dce1b69dca8fe5227bdc67065a9aa7db3f99a#diff-9ad3a8ba51361c2bca2d17d5b7af5dd20320f1a23929f1563984775f0a7eaeebR24).

### DIFF
--- a/backpack/extensions/backprop_extension.py
+++ b/backpack/extensions/backprop_extension.py
@@ -2,7 +2,6 @@
 from __future__ import annotations
 
 import abc
-import warnings
 from abc import ABC
 from typing import Any, Dict, List, Tuple, Type, Union
 
--- a/backpack/extensions/backprop_extension.py
+++ b/backpack/extensions/backprop_extension.py
@@ -60,7 +60,10 @@ class BackpropExtension(ABC):
             AssertionError: if fail_mode is not valid
         """
         if fail_mode not in (FAIL_WARN, FAIL_ERROR, FAIL_SILENT):
-            raise AssertionError(f"no valid fail mode: {fail_mode}")
+            raise AssertionError(
+                f"Invalid failure mode: {fail_mode}."
+                f" Should be one of {(FAIL_WARN, FAIL_ERROR, FAIL_SILENT)}"
+            )
         self.saved_quantities: SavedQuantities = SavedQuantities()
         self.savefield: str = savefield
         self.__module_extensions: Dict[Type[Module], ModuleExtension] = module_exts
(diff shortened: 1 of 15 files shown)


f-dangel:
Hey Fred, I skimmed through your changes:
1. The failing test checks if the result in `batch_grad` for a BN layer in train mode sums to `grad`. Working with `batch_grad` is 'okay' in this case because it's not interpreted as per-sample gradients. We could either revert the default for first-oder fail mode, or adapt the test to use `BatchGrad(fail_mode-"WARNING")`. I would currently favor to revert the default (as this also does not trigger a version bump, and fixes 2.).
2. The RTD example with the custom ResNet fails for similar reasons as in 1.
3. Can you `pip install --upgrade && make black` to update the formatting?

Happy to review or discuss!

PR #256, opened 2022-05-17
[REF] Changed cross entropy loss to NLL base

jabader97:


f-dangel:
Summary:
- Replace shape check of samples in main library with test
- Implement CELoss via NLLLoss
- Add `retain_grad=True` for autograd computation of sampled gradients (for MSELoss, it worked without `retain_graph`)

PR #317, opened 2023-09-05
fix: ScaleModule for DiagHessian.

### LINKED ISSUE #316: Missing implementation of supported layers for DiagHessian and BatchDiagHessian
There are multiple layers which are specified as [being supported](https://docs.backpack.pt/en/master/supported-layers.html) for second order derivatives that actually do not work when trying to calculate the Hessian diagonal using `backpack-for-pytorch<=1.6.0`.

So far, I've run into this problem with the following layers:
* torch.nn.Identity
* torch.nn.BatchNorm1d, torch.nn.BatchNorm2d, torch.nn.BatchNorm3d
* backpack.custom_module.branching.SumModule
* backpack.custom_module.branching.ScaleModule

This can be tested with a script such as the following:
```python
import torch
from backpack import backpack, extend
from backpack.extensions import DiagHessian, BatchDiagHessian
from backpack.custom_module.branching import Parallel, SumModule

model = extend(
    torch.nn.Sequential(
        *[
            torch.nn.Conv2d(3, 16, kernel_size=(3, 3)),
            Parallel(
                torch.nn.Identity(), torch.nn.BatchNorm2d(16), merge_module=SumModule()
            ),
            torch.nn.AdaptiveAvgPool2d(output_size=1),
            torch.nn.Flatten(),
            torch.nn.Linear(16, 2),
        ]
    ).cuda()
)
criterion = extend(torch.nn.CrossEntropyLoss())

batch = torch.randn((2, 3, 8, 8)).cuda()
target = torch.tensor([[1.0, 0.0], [0.0, 1.0]]).cuda()


model.zero_grad()
loss = criterion(model(batch), target)

with backpack(DiagHessian(), BatchDiagHessian()):
    loss.backward()

hessian_diag = torch.cat(
    [p.diag_h.view(-1) for p in model.parameters()], dim=1
)
hessian_diag_batch = torch.cat(
    [p.diag_h_batch.view(batch.shape[0], -1) for p in model.parameters()], dim=1
)
```
I'm guessing that these require independent fixes, but think it is a good idea to collect all layers with missing support summarised here.

hlzl:
Partially fixes #316 . Not sure if `hessian_is_zero` should always return `True`.

### DIFF
--- a/backpack/core/derivatives/scale_module.py
+++ b/backpack/core/derivatives/scale_module.py
@@ -23,3 +23,7 @@ class ScaleModuleDerivatives(BaseDerivatives):
             return mat
         else:
             return mat * module.weight
+
+    def hessian_is_zero(self, module: ScaleModule) -> bool:
+        """Second derivative of a scalar is always = 0."""
+        return True
--- a/backpack/extensions/secondorder/diag_hessian/__init__.py
+++ b/backpack/extensions/secondorder/diag_hessian/__init__.py
@@ -22,6 +22,7 @@ from torch.nn import (
     CrossEntropyLoss,
     Dropout,
     Flatten,
+    Identity,
     LeakyReLU,
     Linear,
     LogSigmoid,
@@ -36,6 +37,7 @@ from torch.nn import (
 )
 
 from backpack.custom_module.pad import Pad
+from backpack.custom_module.scale_module import ScaleModule
 from backpack.custom_module.slicing import Slicing
 from backpack.extensions.secondorder.base import SecondOrderBackpropExtension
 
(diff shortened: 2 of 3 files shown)


hlzl:
Commit 48f03e92e12fd97970a78335fb645ac5ee9a77f2 tries to actually compute the diagonal elements of the Hessian for the batch norm layer.

If one of you could have a quick look at the commits to see if they make any sense, would really appreciate it. @f-dangel @fKunstner 

Thank you!

f-dangel:
Hi,

just wanted to let you know I read your message above. Please don't expect any reaction before the ICLR deadline (Sep 28)

PR #336, opened 2024-11-12
[ADD] `AutomaticDerivatives` for autograd-based support of new layers

f-dangel:
This PR adds an experimental `AutomaticDerivatives` class, which implements BackPACK's core functionality given a module's forward function. This simplifies supporting new layers at the cost of reduced performance. Currently, `AutomaticDerivatives` only implements VJPs w.r.t. layer inputs and parameters. This is enough to support the most common first- and second-order extensions.

A future PR will add an example to demonstrate how to use `AutomaticDerivatives` to add support for a new layer to a specific extension.


f-dangel:
@fKunstner could you have quick look, or give feedback whether you think this feature makes sense?

